PR #10, opened 2021-07-21
add inventory_codes_packing_costs CRUD

miramanz:
DDL: new table: inventory_codes_packing_costs {inventory_code_id,commodity_id,packing-cost(decimal), unique key = commodity_id,inventory_code_id}

UI: User should edit these as an inline editing grid on the edit inventory_code screen. The view inventory code screen should also , as inline grid, its inventory_codes_packing_costs


lib/masterfiles/repositories/fruit_repo.rb
```diff
@@ -96,5 +96,40 @@ def find_rmt_class_by_grade(grade_id)
...
+
+    def sync_inventory_packing_costs(inventory_code_id)
+      select_values(:commodities, :id).each do |commodity_id|
+        args = { commodity_id: commodity_id,
+                 inventory_code_id: inventory_code_id }
+        next if exists?(:inventory_codes_packing_costs, args)
+
+        DB[:inventory_codes_packing_costs].insert(args)
+      end
+    end
```
jcsjcs, line 108: You can do this with a single SQL call:
```ruby
<<~SQL
  INSERT INTO inventory_codes_packing_costs (inventory_code_id, commodity_id)
  SELECT #{inventory_code_id}, commodities.id
  FROM commodities
  WHERE NOT EXISTS (SELECT id 
    FROM inventory_codes_packing_costs 
    WHERE inventory_code_id = #{inventory_code_id} 
    AND commodity_id = commodities.id)
SQL
```

PR #11, opened 2021-07-21
add production client rule: run_cache_legacy_data_fields

miramanz:


lib/client_rules/client_production_rules.rb
```diff
@@ -23,7 +23,8 @@ class ClientProductionRules < BaseClientRules # rubocop:disable Metrics/ClassLen
             capture_product_setup_class: false,
             link_target_markets_to_target_customers: false,
             require_packaging_bom: false,
-            capture_batch_number_for_pallets: true },
+            capture_batch_number_for_pallets: true,
+            run_cache_legacy_data_fields: %i[pc_code track_indicator_code] },
```
jcsjcs, line 28: Shouldn't this be
```ruby
run_cache_legacy_data_fields: []
```
For all clients except `kr`?

PR #12, opened 2021-07-21
add feature to update phyto_data from inspection

FritzViljoen:


lib/finished_goods/repositories/govt_inspection_repo.rb
```diff
@@ -257,6 +257,16 @@ def update_govt_inspection_pallet(id, res)
...
+    def update_govt_inspection_otmc_results(govt_inspection_sheet_id)
+      pallet_ids = select_values(:govt_inspection_pallets, :pallet_id, govt_inspection_sheet_id: govt_inspection_sheet_id)
+      QualityApp::OrchardTestRepo.new.update_otmc_results(pallet_ids)
+    end
+
+    def update_govt_inspection_phyto_data(govt_inspection_sheet_id)
+      pallet_ids = select_values(:govt_inspection_pallets, :pallet_id, govt_inspection_sheet_id: govt_inspection_sheet_id)
+      QualityApp::OrchardTestRepo.new.update_load_phyto_data(pallet_ids)
+    end
+
```
jcsjcs, line 269: Move these methods to the `OrchardTestRepo` and call that repo from the interactor's `update_otmc` method.

Same applies to the two methods in `LoadRepo` and their calling code.

FritzViljoen: pushed changes

PR #13, opened 2021-07-21
added feature skip pallet_verification passed check before adding pal…

FritzViljoen:
…let to govt_inspection_sheet


lib/client_rules/client_fg_rules.rb
```diff
@@ -16,6 +16,8 @@ class ClientFgRules < BaseClientRules # rubocop:disable Metrics/ClassLength
             max_pallets_on_load: 96,
             use_inspection_destination_for_load_out: true,
             use_continuous_govt_inspection_sheets: false,
+            pallet_verification_not_required_for_inspection: false,
+            pallet_weight_not_required_for_inspection: false,
```
jcsjcs, line 20: Change these to positives:
```ruby
pallet_verification_required_for_inspection: true,
pallet_weight_required_for_inspection: true,
```
And use this pattern below:
```ruby
    def can_inspect_without_pallet_weight?(explain: false)
      return 'Can inspections take place without first weighing the pallet?' if explain

      !setting(:pallet_weight_required_for_inspection)
    end
```

FritzViljoen: Applied changes and pushed.

PR #16, opened 2021-07-22
standard_pack_codes

miramanz:
- drop rmt_container_type_id and rmt_container_material_type_id
- add rmt_container_material_owner_id

product_setups
- drop rmt_container_material_owner_id

run_cache
- get rmt_container_material_owner_id from product_setup's standard_pack_code


lib/masterfiles/validations/standard_pack_code_schema.rb
```diff
@@ -12,17 +12,12 @@ class StandardPackContract < Dry::Validation::Contract
...
     end
 
-    rule(:rmt_container_type_id, :bin) do
-      key.failure 'must be filled in' if values[:bin] && !values[:rmt_container_type_id]
-    end
-
-    rule(:rmt_container_material_type_id, :bin) do
-      key.failure 'must be filled in' if values[:bin] && !values[:rmt_container_material_type_id]
+    rule(:rmt_container_material_owner_id, :bin) do
+      key.failure 'must be filled in' if values[:bin] && !values[:rmt_container_material_owner_id]
```
jcsjcs, line 20: This rule should be removed, it is not true that **all** biin standard packs must have a material owner.

PR #28, opened 2021-08-02
Resolve cultivar_id - get value from run

miramanz:
Summary of this change:

Resolve cultivar_id 
- get value from run
- if null get product setup template's value
-  if null and marketing variety's cultivars is not more than 1 get marketing variety's value
-  else leave it as null

CHANGELOG entry:

### Added
- .
### Changed
-  How cultivar_id is resolved for run cache and in turn carton_labels
- 
### Fixed
- .

Production | Runs | List Production Runs
- execute run 

Production | In Progress | Active Setups
- print carton label


lib/production/repositories/production_run_repo.rb
```diff
@@ -453,6 +453,20 @@ def setup_data_for(product_setup_id)
       rec
     end
 
+    def resolve_setup_cultivar_id(product_setup_id)
+      args = DB[:product_setups].where(id: product_setup_id).select(:product_setup_template_id, :marketing_variety_id).first
+      cultivar_id = DB[:product_setup_templates].where(id: args[:product_setup_template_id]).get(:cultivar_id)
+      return cultivar_id unless cultivar_id.nil_or_empty?
```
jcsjcs, line 459: Change `nil_or_empty?` to just `nil?`. `.get` will never return an empty string from an integer field.

lib/production/repositories/production_run_repo.rb
```diff
@@ -453,6 +453,20 @@ def setup_data_for(product_setup_id)
...
+
+      return nil unless only_one_marketing_variety_cultivar?(args[:marketing_variety_id])
+
+      DB[:marketing_varieties_for_cultivars].where(id: args[:marketing_variety_id]).get(:cultivar_id)
+    end
+
+    def only_one_marketing_variety_cultivar?(marketing_variety_id)
+      DB[:marketing_varieties_for_cultivars].where(marketing_variety_id: marketing_variety_id).select_map(:cultivar_id).uniq.count == 1
+    end
```
jcsjcs, line 468: You could reduce the two queries above to just one by getting the results from `select_map` and then applying `.first` if there is only one item in the array.

PR #30, opened 2021-08-04
- Create new reworks run to scrap cartons

miramanz:
Summary of this change:
Create new reworks run to scrap cartons - only cartons that does not reference pallet_sequences (i.e. carton.pallet_sequence_id is null)
User simply pastes a list of carton_label ids (or search and then select from grid- not that important for now). Then normal reworks logging stuff- remember to set carton-scrapped and scrapped_at

CHANGELOG entry:

### Added
- applies_to_cartons to scrap_reasons table and CRUD
- Scrap Cartons and Unscrap Cartons reworks run type
- 
### Changed
- .
### Fixed
- .

Production | Reworks | Scrap Cartons
Production | Reworks | Unscrap Cartons
- User pastes a list of carton_label ids


lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -717,8 +731,11 @@ def create_reworks_run_record(attrs, reworks_action, changes) # rubocop:disable
       res = validate_reworks_run_params(attrs)
       return validation_failed_response(res) if res.failure?
 
-      return create_scrap_bin_reworks_run(res) if repo.find_reworks_run_type(attrs[:reworks_run_type_id])[:run_type] == AppConst::RUN_TYPE_SCRAP_BIN
-      return create_unscrap_bin_reworks_run(res) if repo.find_reworks_run_type(attrs[:reworks_run_type_id])[:run_type] == AppConst::RUN_TYPE_UNSCRAP_BIN
+      reworks_run_type = repo.find_reworks_run_type(attrs[:reworks_run_type_id])[:run_type]
+      return create_scrap_bin_reworks_run(res) if reworks_run_type == AppConst::RUN_TYPE_SCRAP_BIN
+      return create_unscrap_bin_reworks_run(res) if reworks_run_type == AppConst::RUN_TYPE_UNSCRAP_BIN
+      return create_scrap_carton_reworks_run(res) if reworks_run_type == AppConst::RUN_TYPE_SCRAP_CARTON
+      return create_unscrap_carton_reworks_run(res) if reworks_run_type == AppConst::RUN_TYPE_UNSCRAP_CARTON
```
jcsjcs, line 739: Just as a matter of interest (not a requirement to change the code), you could implement this logic using a different pattern:
```ruby
reworks_run_type = repo.get(:reworks_run_types, attrs[:reworks_run_type_id], :run_type)
meth = {
  AppConst::RUN_TYPE_SCRAP_BIN => :create_scrap_bin_reworks_run
  AppConst::RUN_TYPE_UNSCRAP_BIN => :create_unscrap_bin_reworks_run
  AppConst::RUN_TYPE_SCRAP_CARTON => :create_scrap_carton_reworks_run
  AppConst::RUN_TYPE_UNSCRAP_CARTON => :create_unscrap_carton_reworks_run
}[reworks_run_type]
return send(meth, res) unless meth.nil?
```
- This allows you to just change the contents of the hash if you need to handle more or less run types.
- The hash could also be a constant defined outside the method, so the logic would become:
```ruby
reworks_run_type = repo.get(:reworks_run_types, attrs[:reworks_run_type_id], :run_type)
meth = SCRAP_TYPE_CREATE_METHODS[reworks_run_type]
return send(meth, res) unless meth.nil?
```

lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -728,36 +745,55 @@ def create_reworks_run_record(attrs, reworks_action, changes) # rubocop:disable
...
+    end
+
+    def resolve_reworks_run_attrs(params)
+      {
+        user: params[:user],
+        reworks_run_type_id: params[:reworks_run_type_id],
+        scrap_reason_id: params[:scrap_reason_id],
+        remarks: params[:remarks],
+        pallets_selected: repo.array_for_db_col(params[:pallets_selected]),
+        pallets_affected: repo.array_for_db_col(params[:pallets_selected]),
```
jcsjcs, line 793: If these attrs are used by the normal `create_` or `update_` methods, you do not need to call `array_for_db_col`, just supply an array and `BaseRepo` will handle it for you.

lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -2026,9 +2064,32 @@ def validate_production_runs(reworks_run_type, params) # rubocop:disable Metrics
       OpenStruct.new(success: true, instance:  attrs)
     end
 
+    def validate_carton_labels(reworks_run_type, carton_labels) # rubocop:disable Metrics/AbcSize
+      carton_labels = carton_labels.split(/\n|,/).map(&:strip).reject(&:empty?)
+      carton_labels = carton_labels.map { |x| x.gsub(/['"]/, '') }
```
jcsjcs, line 2069: Try not to modify the input parameter directly, you never know how the calling code could use it in future:
```ruby
def validate_carton_labels(reworks_run_type, cartons_string) # rubocop:disable Metrics/AbcSize
      carton_labels = cartons_string.split(/\n|,/).map(&:strip).reject(&:empty?)
      carton_labels.map! { |x| x.gsub(/['"]/, '') }
```

lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -2026,9 +2064,32 @@ def validate_production_runs(reworks_run_type, params) # rubocop:disable Metrics
...
+
+      invalid_carton_labels = carton_labels.reject { |x| x.match(/\A\d+\Z/) }
+      return OpenStruct.new(success: false, messages: { pallets_selected: ["#{invalid_carton_labels.join(', ')} must be numeric"] }, pallets_selected: carton_labels) unless invalid_carton_labels.nil_or_empty?
+
+      existing_carton_labels = repo.carton_labels_exists?(carton_labels, where: { pallet_sequence_id: nil, scrapped_sequence_id: nil  })
+      missing_carton_labels = (carton_labels - existing_carton_labels.map(&:to_s))
+      return OpenStruct.new(success: false, messages: { pallets_selected: ["#{missing_carton_labels.join(', ')} doesn't exist"] }, pallets_selected: carton_labels) unless missing_carton_labels.nil_or_empty?
```
jcsjcs, line 2076: Isn't this just the equivalent of:
```ruby
DB[:carton_labels]
.where(id: carton_labels)
.exclude(pallet_sequence_id: nil)
.exclude(scrapped_sequence_id: nil)
.select(:id)
```
or am I missing something?

OK, having looked at the implementation of `carton_label_exists?`, I see that you are looking for verified cartons.
So maybe use the query above and also something like this to pick up labels that have not been verified:
```sql
SELECT id FROM carton_labels
WHERE id IN (carton_labels)
AND NOT EXISTS(SELECT id FROM cartons WHERE carton_label_id = carton_labels.id)
```

lib/production/repositories/reworks_repo.rb
```diff
@@ -1146,5 +1146,13 @@ def pallet_requires_material_owner?(pallet_number)
...
     end
+
+    def carton_labels_exists?(carton_labels, where: {})
+      DB[:carton_labels]
+        .join(:cartons, carton_label_id: :id)
+        .where(Sequel[:carton_labels][:id] => carton_labels)
+        .where(where)
+        .select_map(Sequel[:carton_labels][:id])
+    end
   end
```
jcsjcs, line 1157: Apart from the comments above, this method should not end in `?` as it does not return a boolean.
Also it should be renamed because what it does is return verified carton label ids from a set that meet certain criteria, not check for existence.

lib/production/ui_rules/reworks_run_rule.rb
```diff
@@ -57,7 +67,7 @@ def set_show_fields # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplex
       fields[:scrap_reason_id] = { renderer: :label,
                                    with_value: scrap_reason_id_label,
                                    caption: 'Scrap Reason',
-                                   hide_on_load: @rules[:scrap_pallet] || @rules[:scrap_bin] ? false : true }
+                                   hide_on_load: @rules[:scrap_pallet] || @rules[:scrap_bin] || @rules[:scrap_carton] ? false : true }
```
jcsjcs, line 70: This pattern occurs a few times. Maybe create a method named `scrapping?` which checks the three rules and returns true or false.

lib/production/ui_rules/reworks_run_rule.rb
```diff
@@ -151,23 +164,35 @@ def common_fields # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexit
...
                      end
-      text_area_caption = @rules[:bin_run_type] ? 'Bins' : 'Pallets'
-      scrap_reason = @rules[:scrap_bin] ? MasterfilesApp::QualityRepo.new.for_select_scrap_reasons(where: { applies_to_bins: true }) : MasterfilesApp::QualityRepo.new.for_select_scrap_reasons(where: { applies_to_pallets: true })
+      text_area_caption = if @rules[:bin_run_type]
+                            'Bins'
+                          elsif @rules[:carton_run_type]
+                            'Cartons'
+                          else
+                            'Pallets'
+                          end
```
jcsjcs, line 173: `text_area_caption` can be turned into a method.

jcsjcs (changes requested):

lib/production/interactors/reworks_run_interactor.rb
miramanz: Noted

lib/production/interactors/reworks_run_interactor.rb
miramanz: scrap cartons is only for cartons that does not reference pallet_sequences (i.e. carton.pallet_sequence_id is null). So only cartons where pallet_sequence_id is null and scrapped_sequence_id is null for scrapped cartons.
Cartons referenced to a pallet_sequence can already be scrapped (viewed via pallet_sequence grid).

jcsjcs: OK, so then you just do this to get the `missing_carton_labels`:
```ruby
DB[:cartons]
.where(carton_label_id: carton_labels)
.exclude(pallet_sequence_id: nil, scrapped_sequence_id: nil)
.select(:carton_label_id)
```
That generates the following SQL:
```sql
SELECT "carton_label_id"
FROM "cartons"
WHERE (("carton_label_id" IN (1, 2, 3, 4))
AND (("pallet_sequence_id" IS NOT NULL) OR ("scrapped_sequence_id" IS NOT NULL)))
```

PR #37, opened 2021-08-12
Feature/auto create destination for pallet buildup

lmatoti:
Summary of this change:

Pallet Buildups

Build a pallet by transferring cartons from other pallets

CHANGELOG entry:
Added

Ability to not scan a destination pallet and create it on the fly instead.

Changed

Fixed

.

RMD | Buildups | Pallet Buildup

1. New field 'Auto Create Destination Pallet': click it, 'Destination Pallet' is removed from the screen i.e. turns destination pallet on and off
2. Scan source pallets.
3. Scan cartons
4. OnComplete: Destination pallet will be created and scanned cartons moved to it.


lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -4,12 +4,12 @@ module FinishedGoodsApp
   class BuildupsInteractor < BaseInteractor # rubocop:disable Metrics/ClassLength
     def process_to_rejoin(params) # rubocop:disable Metrics/AbcSize
       dest = params[:pallet_number]
-      source = params.select { |k, v| !v.nil_or_empty? && !k.to_s.include?('_scan_field') && !k.to_s.include?('qty_to_move') && !k.to_s.include?('pallet_number') }.values
+      source = params.select { |k, v| !v.nil_or_empty? && !k.to_s.include?('_scan_field') && !k.to_s.include?('qty_to_move') && !k.to_s.include?('pallet_number') && !k.to_s.include?('auto_create_destination_pallet') }.values
```
jcsjcs, line 7: It seems you want only the values of all params with key `:pnn` (p followed by one or more digits) that are not empty or nil.
This should do it and be future-proof if you add other keys to the params:
```ruby
params.select { |k,v| k.to_s.match(/^p\d+$/) && !v.to_s.empty? }.values.compact
```
Another way would be to get the values for relevant keys, ignoring empty strings and removing nils, like this:
```ruby
params.values_at(*(1..10).map { |n| "p#{n}".to_sym }).grep_v('').compact

# Or break into two so it's easier to read:
pallet_keys = (1..10).map { |n| "p#{n}".to_sym }
params.values_at(*pallet_keys).grep_v('').compact
```
You could also change 10 to 20 so this will work even if the number of pallets is later changed from 10.

lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -4,12 +4,12 @@ module FinishedGoodsApp
...
       repo.get_process_to_rejoin(dest, source, @user.user_name)
     end
 
     def process_to_cancel(params) # rubocop:disable Metrics/AbcSize
-      pallets = params.select { |k, v| !v.nil_or_empty? && !k.to_s.include?('_scan_field') && !k.to_s.include?('qty_to_move') }.values
+      pallets = params.select { |k, v| !v.nil_or_empty? && !k.to_s.include?('_scan_field') && !k.to_s.include?('qty_to_move') && !k.to_s.include?('auto_create_destination_pallet') }.values
```
jcsjcs, line 12: Same comment as above applies here.

lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -18,7 +18,10 @@ def process_to_cancel(params) # rubocop:disable Metrics/AbcSize
     end
 
     def buildup_pallet(params) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      return validation_failed_response(OpenStruct.new(messages: { pallet_number: ['is required'] }))  if params[:auto_create_destination_pallet] == 'f' && params[:pallet_number].nil_or_empty?
```
jcsjcs, line 21: You can use a dry validation contract on these parameters that checks that pallet number is needed when auto-create is false. The validation can also exclude the `_scan_fields` for you.

lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -91,15 +94,28 @@ def delete_pallet_buildup(id)
...
+      repo.transaction do # rubocop:disable Metrics/BlockLength
         pallet_buildup = pallet_buildup(id)
         unless pallet_buildup.completed
-          dest_pallet = repo.get_value(:pallets, :id, pallet_number: pallet_buildup.destination_pallet_number)
+          if !pallet_buildup.destination_pallet_number.nil?
+            dest_pallet = repo.get_value(:pallets, :id, pallet_number: pallet_buildup.destination_pallet_number)
+          else
+            creator_ctn_label = pallet_buildup.cartons_moved.first[1][0]
+            carton_id, orig_seq = repo.select_values(:cartons, %i[id pallet_sequence_id], carton_label_id: creator_ctn_label).flatten
+            res = MesscadaApp::CreatePalletFromCarton.new(@user, carton_id, 1, nil, false).call
```
jcsjcs, line 106: This should work without the default parameters?
```ruby
res = MesscadaApp::CreatePalletFromCarton.new(@user, carton_id, 1).call
```

lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -91,15 +94,28 @@ def delete_pallet_buildup(id)
...
+            dest_pallet = res.instance[:pallet_id]
+          end
+
           pallet_buildup.cartons_moved.each do |_k, v|
             v.each do |cl_id|
-              res = MesscadaApp::TransferCarton.call(ProductionApp::ProductionRunRepo.new.find_carton_by_carton_label_id(cl_id)[:id], dest_pallet)
-              return res unless res.success
+              unless cl_id == creator_ctn_label
+                res = MesscadaApp::TransferCarton.call(prod_run_repo.find_carton_by_carton_label_id(cl_id)[:id], dest_pallet)
+                raise res.message unless res.success
```
jcsjcs, line 117: Rather raise a `Crossbeams::InfoError` here.

jcsjcs (changes requested):

lmatoti:
pallet_buildups
changes implemented as per code review

PR #46, opened 2021-08-20
re-execute run after change_run_* reworks_runs

miramanz:
Summary of this change:
re-execute run after change_run_orchard and change_run_cultivar reworks_run types for production_runs where labeling is true

CHANGELOG entry:

### Added
- update reconfiguring and setup_complete attributes to production_run
- re-execute run after change for production_runs where labeling is true

Production | Reworks | Change Run Orchard
Production | Reworks | Change Run Cultivar


lib/production/services/change_run_cultivar.rb
```diff
@@ -17,15 +17,17 @@ def call
       res = change_run_cultivar
       raise Crossbeams::InfoError, unwrap_failed_response(res) unless res.success
 
+      ReExecuteRun.call(production_run_id, user_name) if params[:labeling]
```
jcsjcs, line 20: This should be inside the same transaction as the updates.

lib/production/services/change_run_orchard.rb
```diff
@@ -17,15 +17,17 @@ def call
       res = change_run_orchard
       raise Crossbeams::InfoError, unwrap_failed_response(res) unless res.success
 
+      ReExecuteRun.call(production_run_id, user_name) if params[:labeling]
```
jcsjcs, line 20: This too should be inside the main transaction.

jcsjcs (changes requested):

PR #59, opened 2021-09-16
Replace  the subquery in the join with a subquery in the select

PatieC:
Summary of this change:

View was returning duplicates as result of the subquery in the join.

The subquery was removed from join.
Two subqueries to retrieve repacked_to_pallet_number and repacked_to_pallet_id were added in the select clause.


jcsjcs:
Re-created as a hotfix.

PR #70, opened 2021-09-30
Create feature on loads menu, visible for shipped loads only

miramanz:
Summary of this change:
Create feature on loads menu, visible for shipped loads only, called 'force_shipped_date'
Actionable for users with load  'can_ship'  user permission

### Added
- . a feature on loads menu, visible for shipped loads only, called 'force_shipped_date'

Finished Goods | Dispatch | List Loads | force shipped date

To change user permissions
Development | Masterfiles | Users | User permissions


grid_definitions/lists/loads.yml
```diff
@@ -29,6 +29,12 @@
   :title: Re-send HCS EDI
   :popup: true
   :hide_if_false: shipped,on_order
+- :url: "/finished_goods/dispatch/loads/$:id$/force_shipped_date"
+  :text: force shipped date
+  :icon: edit
+  :title: Edit shipped date
+  :popup: true
+  :hide_if_false: shipped
```
jcsjcs, line 37: Please use the `has_permission` option here so that it is only visible to people with the `can_ship` permission.

routes/finished_goods/dispatch.rb
```diff
@@ -420,6 +420,38 @@ class Nspack < Roda
         show_partial_or_page(r) { FinishedGoods::Orders::Order::New.call(form_values: form_values) }
       end
 
+      r.on 'force_shipped_date' do
```
jcsjcs, line 423: Create a method in your interactor to check `UserPermissions` for `can_ship` here.

lib/finished_goods/interactors/load_interactor.rb
```diff
@@ -349,6 +349,35 @@ def load_entity(id)
       repo.find_load(id)
     end
 
+    def force_shipped_date(load_id, params)
```
jcsjcs, line 352: This method is miss-named.
You're just fetching the date.

lib/finished_goods/interactors/load_interactor.rb
```diff
@@ -349,6 +349,35 @@ def load_entity(id)
...
+      can_ship_load = Crossbeams::Config::UserPermissions.can_user?(@user, :load, :can_ship) unless @user&.permission_tree.nil?
```
jcsjcs, line 353: This check can be moved to the `confirm_force_shipped_date` method and also called from the route.

jcsjcs (changes requested):

PR #74, opened 2021-10-01
Feature: Allocate work order items to product_resource_allocations

miramanz:
Summary of this change:
Feature: Allocate work order items to product_resource_allocations

### Changed
- . When user selects a product_setup or packing_specification , do a dynamic lookup to find work order items. If only one work_order_item is found, allocate it (and display on grid) automatically, else populate possible values in grid
- Cascade work_order_item to carton labels and pallet_sequences

Production | Runs | List Production Runs | Allocate Product Setups


lib/production/repositories/production_run_repo.rb
```diff
@@ -331,6 +331,48 @@ def allocate_packing_specification(product_resource_allocation_id, packing_speci
       success_response("Allocated #{packing_specification_item_code}", packing_specification_item_code: packing_specification_item_code)
     end
 
+    def automatically_allocate_work_order_item(product_resource_allocation_id)
+      items = work_order_items_for(product_resource_allocation_id)
+      return if items.count > 1
```
jcsjcs, line 336: Please make this easier to understand by adding a comment
```ruby
# Only auto-allocate if there is zero or exactly one matching item (otherwise the user has to choose the match):
```

lib/production/repositories/production_run_repo.rb
```diff
@@ -331,6 +331,48 @@ def allocate_packing_specification(product_resource_allocation_id, packing_speci
...
+
+      allocate_work_order_item(product_resource_allocation_id, items[0])
```
jcsjcs, line 338: Please use `items.first` rather than `items[0]`.

lib/production/repositories/production_run_repo.rb
```diff
@@ -331,6 +331,48 @@ def allocate_packing_specification(product_resource_allocation_id, packing_speci
...
+    end
+
+    def work_order_items_for(product_resource_allocation_id)
+      setup_id, target_cust_id = DB[:product_resource_allocations]
+                                 .where(id: product_resource_allocation_id)
+                                 .select(:product_setup_id, :target_customer_party_role_id)
```
jcsjcs, line 344: The `select` is not needed when you use `get`.

lib/production/repositories/production_run_repo.rb
```diff
@@ -331,6 +331,48 @@ def allocate_packing_specification(product_resource_allocation_id, packing_speci
...
+                                 .get(%i[product_setup_id target_customer_party_role_id])
+      extra_conditions = target_cust_id.nil? ? '' : " AND marketing_orders.customer_party_role_id = #{target_cust_id}"
+      query = <<~SQL
+        SELECT DISTINCT fn_work_order_item_code(work_order_items.id) AS code
+        FROM work_order_items
+        JOIN work_orders ON work_orders.id = work_order_items.work_order_id
+        LEFT JOIN marketing_orders ON marketing_orders.id = work_orders.marketing_order_id
+        WHERE work_order_items.product_setup_id = #{setup_id} #{extra_conditions}
+      SQL
+      DB[query].select_map(:code)
```
jcsjcs, line 354: Please change to:
```ruby
DB[query, setup_id]
```
This should ideally be rewritten as something like:
```ruby
ds = DB[:work_order_items]
.left_join()
.where(setup_id: setup_id)
ds = ds.where(extra) if extra
ds.select_map(:code)
```

jcsjcs (changes requested):

PR #86, opened 2021-10-11
Work order item fulfillments

miramanz:
Summary of this change:
Work order item fulfillments

### Added
- . wo_fulfillment_pallet_warning_level  and work_order_managers app constants
- pallet_fulfillment_warning_level to work_order_items table
- wo_fulfillment_queue and processor that updates work_order_items.quantity_produced (total quantity of cartons across all pallet_sequences for each work_order_item- matching on season_id). It also sends an email to work_order_managers when work_order_item is within wo_fulfillment_pallet_warning_level
- work order item color rules

Production | Orders | Work Orders | Work Order List | items


scripts/process_wo_fulfillment_queue.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  def run # rubocop:disable Metrics/AbcSize
+    @repo = DevelopmentApp::UserRepo.new
+    @queue_ids = repo.select_values(:wo_fulfillment_queue, :id)
+    @email_list = repo.email_addresses(user_email_group: AppConst::EMAIL_WORK_ORDER_MANAGERS)
+
+    DB.transaction do
+      unless queue_ids.empty?
+        work_order_item_ids_for_fulfillment(queue_ids).each do |work_order_item_id|
+          @work_order_item_code = DB.get(Sequel.function(:fn_work_order_item_code, work_order_item_id))
```
jcsjcs, line 31: This can be part of the query in `work_order_item_ids_for_fulfillment` so that you don't have to run a query here for every loop item.

scripts/process_wo_fulfillment_queue.rb
```diff
@@ -0,0 +1,114 @@
...
+      .where(season_id: DB[:work_order_items]
+                          .join(:work_orders, id: Sequel[:work_order_items][:work_order_id])
+                          .join(:marketing_orders, id: :marketing_order_id)
+                          .where(Sequel[:work_order_items][:id] => work_order_item_id)
+                          .get(:season_id))
+      .sum(:carton_quantity).to_i
+  end
+
+  def pallet_warning_level_reached?(work_order_item_id)
+    DB.get(Sequel.function(:fn_woi_pallets_outstanding, work_order_item_id)).< AppConst::WO_FULFILLMENT_PALLET_WARNING_LEVEL
```
jcsjcs, line 67: Remove the `.` before `<` and replace with a space.

scripts/process_wo_fulfillment_queue.rb
```diff
@@ -0,0 +1,114 @@
...
+      Loops through the wo_fullfillment_queue records and
+      1. calculates the total quantity of cartons across all pallet_sequences for each work_order_item matching on season_id
+      2. updates work_order_item.quantity_produced
+      3. sends an email for each work order item within wo_fulfillment_pallet_warning_level
+
+      Results:
+      --------
+      output:
+      queued wo_fulfillment_queue ids = #{queue_ids}
+
```
jcsjcs, line 107: Perhaps also log which ids were close to fulfillment - where an email was sent.

jcsjcs (changes requested):

scripts/process_wo_fulfillment_queue.rb
```diff
@@ -0,0 +1,122 @@
...
+  def run # rubocop:disable Metrics/AbcSize
+    @repo = DevelopmentApp::UserRepo.new
+    @email_list = repo.email_addresses(user_email_group: AppConst::EMAIL_WORK_ORDER_MANAGERS)
+
+    @work_order_item_ids = []
+    DB.transaction do
+      @queue_ids = repo.select_values(:wo_fulfillment_queue, :id)
+
+      unless queue_ids.empty?
+        work_order_item_ids_for_fulfillment(queue_ids).each do |work_order_item_id, work_order_item_code|
```
jcsjcs, line 32: Change the block variable `work_order_item_code` to something else (like `woi_code`).
It is confusing (and could lead to accidental bugs) to use the same name as the instance variable.

jcsjcs (changes requested):

PR #133, opened 2021-11-02
Force Offload Bin/Pallet Tripsheet

lmatoti:
Summary of this change:
Force Offload Bin/Pallet Tripsheet
force_offload permission controls if user can force offload tripsheet


lmatoti:
Force Offload Bin/Pallet Tripsheet
force_offload user_permission controls if user can force offload tripsheet

PR #137, opened 2021-11-02
Offload Bins Tripsheet

lmatoti:
Summary of this change:
Offload Bins Tripsheet
Scan tripsheet and location and then offload tripsheet without having to scan individual bins.


lib/raw_materials/interactors/rmt_bin_interactor.rb
```diff
@@ -175,6 +175,28 @@ def force_bin_tripsheet_offload(vehicle_job_id) # rubocop:disable Metrics/AbcSiz
       failed_response(e.message, stock_type_id)
     end
 
+    def offload_bins_trip(vehicle_job_id, location_id) # rubocop:disable Metrics/AbcSize
+      res = validate_bins_tripsheet_to_offload_(vehicle_job_id, location_id)
+      return res unless res.success
+
+      repo.transaction do
+        vehicle_job_units = repo.select_values(:vehicle_job_units, :id, vehicle_job_id: vehicle_job_id)
+        bin_ids = repo.select_values(:vehicle_job_units, :stock_item_id, vehicle_job_id: vehicle_job_id)
```
jcsjcs, line 184: This query should also include the `stock_type_id`.
Both queries could be combined as:
```ruby
ids = repo.select_values(:vehicle_job_units, %i[id stock_item_id], vehicle_job_id: vehicle_job_id, stock_type_id: stock_type_id)
unit_ids = ids.map(&:first)
bin_ids = ids.map(&:last)
```

jcsjcs (changes requested):

PR #141, opened 2021-11-04
add bin asset number field to weigh rmt bins reworks run

miramanz:
Summary of this change:
add bin asset number field to weigh rmt bins reworks run

### Added
- . bin asset number field to weigh rmt bins reworks run

Production | Reworks | Weigh Rmt Bins | New


lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -652,6 +652,17 @@ def resolve_log_reworks_runs_status_and_transaction(reworks_run_type, reworks_ru
 
     def create_reworks_run(reworks_run_type_id, params) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       reworks_run_type = reworks_run_type(reworks_run_type_id)
+
+      if AppConst::RUN_TYPE_WEIGH_RMT_BINS == reworks_run_type
+        res = ReworksWeighRmtBinsContract.new.call(params.merge(pallets_selected: Array(params[:pallets_selected]).reject(&:empty?)))
+        return validation_failed_response(res) if res.failure?
+
+        rmt_bin_id = res[:pallets_selected].nil_or_empty? ? repo.get_id(:rmt_bins, bin_asset_number: res[:bin_asset_number]) : params[:pallets_selected]
```
jcsjcs, line 660: Why don't you only do these three lines if `pallets_selected` is nil?
I assume that despite its name, `pallets_selected` can only contain one value? Otherwise there'd be a problem with the 3rd line.

lib/production/validations/reworks_run_schema.rb
```diff
@@ -259,4 +259,19 @@ module ProductionApp # rubocop:disable Metrics/ModuleLength
...
+
+  class ReworksWeighRmtBinsContract < Dry::Validation::Contract
+    params do
+      required(:reworks_run_type_id).filled(:integer)
+      required(:pallets_selected).maybe(:array).each(:string)
+      optional(:make_changes).maybe(:bool)
+      required(:bin_asset_number).maybe(Types::StrippedString)
+    end
+
+    rule(:pallets_selected, :bin_asset_number) do
```
jcsjcs, line 271: This would be nicer if it was split into two rules - one for both empty and one for both filled.

jcsjcs (changes requested):

PR #159, opened 2021-11-16
Reprint carton labels of pallets that has_indiv_cartons

miramanz:
Summary of this change:
Reprint carton labels of pallets that has_indiv_cartons: users should multi-select the list of cartons to print, then send the command to the printer

### Added
- . Reprint carton labels of pallets that has_indiv_cartons: users should multi-select the list of cartons to print, then send the command to the printer

Production | Reworks | Single Pallet Edit | Reprint Pallet Carton Labels


lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -1858,6 +1858,34 @@ def remove_work_in_progress_lock(reworks_run_type_id, multiselect_list) # ruboco
...
+
+      success_response('ok', res.to_h)
+    end
+
+    def reprint_pallet_carton_labels(carton_ids, params) # rubocop:disable Metrics/AbcSize
+      return failed_response('Carton selection cannot be empty') if carton_ids.nil_or_empty?
+
+      label_name = repo.get(:label_templates, params[:label_template_id], :label_template_name)
+      labels = repo.reworks_run_pallet_seq_print_data_for_cartons(carton_ids)
+      repo.transaction do
```
jcsjcs, line 1874: The `PrintLabel` service does not update the database, so a transaction and log_transaction call is not required here.

jcsjcs (changes requested):

PR #171, opened 2021-11-22
Presort Integration Queue Job

lmatoti:
Summary of this change:
Presort Integration Queue Job
Presort integration events (bin_tipped and bin_created) now use a job queue


lib/raw_materials/jobs/presort_integration_queue_job.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module RawMaterialsApp
+  module Job
+    class PresortIntegrationQueue < BaseQueJob
+      self.maximum_retry_count = 0
+
```
jcsjcs, line 7: This job needs to implement the `single_instance_job` method so that only one job can run at a time.

lib/raw_materials/jobs/presort_integration_queue_job.rb
```diff
@@ -0,0 +1,21 @@
...
+      def run(bin_number, unit, request_path, event)
+        if event == 'create'
+          AppConst::PRESORT_BIN_CREATED_LOG.info("#{request_path}&bin=#{bin_number}&unit=#{unit}")
+          MesscadaApp::PresortBinCreated.call(bin_number, unit)
+        elsif event == 'tipped'
+          AppConst::PRESORT_BIN_TIPPED_LOG.info("#{request_path}&bin=#{bin_number}")
+          MesscadaApp::PresortBinTipped.call(bin_number, unit)
+        end
```
jcsjcs, line 15: There should be an `else` here that raises an exception if the job is called with an `event` value other than `create` or `tipped`.

jcsjcs (changes requested):

lmatoti:
Presort Integration Queue Job
Ensures that only one job can run at a time

PR #175, opened 2021-11-24
tripsheet logs

lmatoti:
Summary of this change:
tripsheet logs


grid_definitions/lists/bin_vehicle_job_units.yml
```diff
@@ -0,0 +1,8 @@
+---
+:dataminer_definition: bin_vehicle_job_units
+:conditions:
+  :standard:
+    - :col: vehicle_job_units.vehicle_job_id
+      :op: =
+      :val: $:vehicle_job_id$
+      :optional: true
```
jcsjcs, line 8: Why is this parameter optional? This doesn't make sense - it is the only param for this condition.
If the tripsheet grid could have a null value for vehicle_job_id then rather use `hide_if_null: id` to the calling list's action.

grid_definitions/lists/vehicle_job_units.yml
```diff
@@ -0,0 +1,8 @@
+---
+:dataminer_definition: vehicle_job_units
+:conditions:
+  :standard:
+    - :col: vehicle_job_units.vehicle_job_id
+      :op: =
+      :val: $:vehicle_job_id$
+      :optional: true
```
jcsjcs, line 8: Why is this parameter optional? This doesn't make sense - it is the only param for this condition.

jcsjcs (changes requested):

lmatoti:
Code review changes made

PR #181, opened 2021-11-26
Extended Deliveries Data Model

lmatoti:
Summary of this change:
Extended Deliveries Data Model


jcsjcs:
There are several problems with this request, so the migration has been written from scratch in https://github.com/NoSoft-SA/nspack/commit/84cb92992eeb3de28712164b40d33583ca61596e

* No `created_at` or `updated_at` columns.
* No uniqueness constraints.
* Main table columns were nullable.
* No audit triggers.
* Some missing columns.
* Down migrations for altered tables did not remove all columns.

PR #199, opened 2021-12-13
make pm_mark a required field on editing pallet_sequences in reworks.

miramanz:
Summary of this change:
make pm_mark a required field on editing pallet_sequences in reworks.

### Changed
- . make pm_mark a required field on editing pallet_sequences in reworks.

Production | Reworks | Single Pallet Edit


lib/production/ui_rules/reworks_run_sequence_rule.rb
```diff
@@ -243,6 +243,7 @@ def edit_sequence_fields # rubocop:disable Metrics/AbcSize
                                ),
                                disabled_options: MasterfilesApp::BomRepo.new.for_select_inactive_pm_marks,
                                caption: 'PKG Mark',
+                               required: true,
```
jcsjcs, line 246: This can only be required if `AppConst::CR_PROD.use_packing_specifications?` is true.

lib/production/validations/reworks_run_schema.rb
```diff
@@ -281,7 +281,7 @@ class SequenceSetupDataContract < Dry::Validation::Contract
       optional(:pm_type_id).maybe(:integer)
       optional(:pm_subtype_id).maybe(:integer)
       required(:target_market_id).maybe(:integer)
-      optional(:pm_mark_id).maybe(:integer)
+      optional(:pm_mark_id).filled(:integer)
```
jcsjcs, line 284: Leave this as `maybe` and add a rule that checks for presence if `AppConst::CR_PROD.use_packing_specifications?` is true.

jcsjcs (changes requested):

jcsjcs:
Implemented as a hotfix.

PR #205, opened 2021-12-16
Feature/delivery bin type and owner

lmatoti:
Summary of this change:
Delivery Bin Type and Material Owner
1. when creating bins after the delivery has been created: bins will inherit the rmt-code and classifications from the delivery
2. Implemented the feature that allows users to capture the container and material on the delivery so that it can be inherited by bins


lib/masterfiles/test/factories/rmt_container_factory.rb
```diff
@@ -11,6 +11,129 @@ def create_rmt_container_material_owner(opts = {})
       DB[:rmt_container_material_owners].insert(opts)
     end
 
+    def create_rmt_code(opts = {})
```
jcsjcs, line 14: Please delete the scaffold comments in this file.

lib/raw_materials/interactors/rmt_delivery_interactor.rb
```diff
@@ -153,13 +155,15 @@ def create_rmt_delivery(params)
...
-    def update_rmt_delivery(id, params) # rubocop:disable Metrics/AbcSize
+    def update_rmt_delivery(id, params) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity:
       res = validate_rmt_delivery_params(params)
       return failed_response(unwrap_failed_response(validation_failed_response(res))) if res.failure? && res.errors.to_h.one? && res.errors.to_h.include?(:season_id)
       return validation_failed_response(res) if res.failure?
 
       repo.transaction do
         repo.update_rmt_delivery(id, res)
+        bin_ids = repo.select_values(:rmt_bins, :id, rmt_delivery_id: id)
+        repo.update(:rmt_bins, bin_ids, rmt_container_type_id: res[:rmt_container_type_id], rmt_material_owner_party_role_id: res[:rmt_material_owner_party_role_id]) if AppConst::CR_RMT.single_bin_type_for_rmt_delivery?
```
jcsjcs, line 166: Both these lines should be inside the `if` - otherwise you will run an extra query fetching bin ids for other clients when it is not required.

jcsjcs (changes requested):

PR #245, opened 2022-01-10
Cosmetic delivery changes

lmatoti:
Summary of this change:
Cosmetic delivery changes — View

    Edit/view delivery: tripsheets:print tripsheet button should always be visible if tripsheet exist
    view delivery:
    2.1 Change view to use an info table instead of labels (use three columns- classifications in third column)
    2.2. Include the following new values inside info table : variant, regime,rmt_code,classifications (only for each value that exists- caption must be the classification's type)

     2.2.1 Do not show these tables if values do not exist (hide table column)

    Edit delivery: show the current read-only labels as an info table with three columns at the top- and include new values listed in 2.2
    Deliveries grid: add the following cols: variant, regime, rmt_code, classification1, classification2, classfications


lib/raw_materials/ui_rules/rmt_delivery_rule.rb
```diff
@@ -30,89 +30,122 @@ def generate_rules # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexi
       form_name 'rmt_delivery'
     end
 
-    def set_show_fields # rubocop:disable Metrics/AbcSize
-      rmt_container_type_id_label = @repo.get(:rmt_container_types, :container_type_code, @form_object.rmt_container_type_id)
-      rmt_owner_label = MasterfilesApp::PartyRepo.new.fn_party_role_name(@form_object.rmt_material_owner_party_role_id)
-      rmt_container_material_type_id_label = @repo.get(:rmt_container_material_types, :container_material_type_code, @form_object.rmt_container_material_type_id)
+    def set_show_fields # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      rules[:col1_data] = {}
+      rules[:col1_labels] = %i[delivery_id sample_bin_positions season farm puc orchard cultivar destination reference_number sample_bins_weighed?]
```
jcsjcs, line 35: Instead of the complication of three tables, please just use `compact_header` with `display_columns` set to 3.
If `find_rmt_delivery_flat` doesn't provide all the columns you need, then build a separate object based on `@form_object` and use the `with_object` option.
Then all you need to do is remove items from the `columns` array if they should not be displayed.
```ruby
cols = %i[delivery_id batch_number season farm puc]
cols.delete(:batch_number) if @form_object.batch_number.nil_or_empty?
rules[:compact_header = compact_header(cols, display_columns: 3}

# OR:
header = @form_object.to_h
header[:something] = 'Some value'
rules[:compact_header = compact_header(cols, display_columns: 3, with_object: header}
```
And in the view:
```ruby
section.add_text rules[:compact_header]
```

lib/raw_materials/views/rmt_delivery/edit.rb
```diff
@@ -59,7 +59,7 @@ def self.call(id, is_update: nil, form_values: nil, form_errors: nil) # rubocop:
 
               print_tripsheet = { text: 'Print Tripsheet',
                                   url: "/rmd/finished_goods/print_tripsheet/#{rules[:vehicle_job_id]}",
-                                  visible: rules[:cancel_delivery_tripheet],
+                                  visible: rules[:print_delivery_tripheet],
```
jcsjcs, line 62: The Tripsheets dropdown button should also have its visible option set.
If there are no tripsheets and none can be created (i.e. the dropdown button would not have any visible items) the whole dropdown button should not appear.

jcsjcs (changes requested):

PR #263, opened 2022-01-26
Bulk Rebin & Pallet updates in reworks

miramanz:
Summary of this change:
Bulk Rebin & Pallet updates in reworks - also copy over from run to rebins: rmt_code_id, colour_percentage (only if null on rebin), actual cold_treatment, actual ripeness treatment

### Changed
- . Bulk Rebin & Pallet updates in reworks - also copy over from run to rebins: rmt_code_id, colour_percentage (only if null on rebin), actual cold_treatment, actual ripeness treatment

Production | Reworks | Bulk Pallet Run Update | New
Production | Reworks | Bulk Rebin Run Update | New

### DIFF
--- a/lib/production/interactors/reworks_run_interactor.rb
+++ b/lib/production/interactors/reworks_run_interactor.rb
@@ -601,8 +601,14 @@ module ProductionApp
         children = repo.select_values(:pallet_sequences, :id, { pallet_number: attrs[:pallets_selected], production_run_id: attrs[:from_production_run_id] })
         before_attrs = production_run_attrs(attrs[:from_production_run_id], production_run(attrs[:from_production_run_id]))
         after_attrs = production_run_attrs(attrs[:to_production_run_id], production_run(attrs[:to_production_run_id]))
-        change_descriptions = { before: production_run_details(attrs[:from_production_run_id]).sort.to_h,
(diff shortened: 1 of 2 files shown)


lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -637,13 +637,18 @@ def resolve_update_selected_input(reworks_run_type, attrs) # rubocop:disable Met
...
+                 puc_id: instance[:puc_id],
+                 orchard_id: instance[:orchard_id],
+                 cultivar_group_id: instance[:cultivar_group_id],
+                 cultivar_id: instance[:cultivar_id],
+                 season_id: instance[:season_id],
+                 rmt_code_id: instance[:rmt_code_id],
+                 actual_cold_treatment_id: instance[:actual_cold_treatment_id],
+                 actual_ripeness_treatment_id: instance[:actual_ripeness_treatment_id] }
+      attrs[:colour_percentage_id] = instance[:colour_percentage_id] unless instance[:colour_percentage_id].nil?
+      attrs
```
jcsjcs, line 651: The run's colour percentage should always be included in the attributes.
But the value should only be applied to the bin if its colour percentage is nil.
The way it is here, the bin's colour percentage will always be replaced.

jcsjcs (changes requested):

PR #274, opened 2022-02-11
Web Bin Weighing And Tipping

lmatoti:
Summary of this change:
Web Bin Weighing And Tipping

CHANGELOG entry:

### Added
- . Bin weighing/tipping web app

Production | Runs | Tip Bin
-. Fill form and submit. Bin will be weighed and tipped.


lib/messcada/interactors/messcada_interactor.rb
```diff
@@ -105,6 +105,22 @@ def update_bin_weights_and_tip_bin(params) # rubocop:disable Metrics/AbcSize
       e.message
     end
 
+    def weigh_and_tip_bin(params) # rubocop:disable Metrics/AbcSize
+      repo.transaction do
```
jcsjcs, line 109: You need to validate the params.

PR #275, opened 2022-02-11
Pallet Buildup

lmatoti:
Summary of this change:
Pallet Buildup
No longer stores the carton used to create the pallet in pallet_sequence.scanned_from_carton_id, stores it in buildup_scanned_carton_id instead


jcsjcs (changes requested):
Please change the `initialize` methods for both `CreatePalletFromCarton` and `NewPalletSequence` and all calls to those services where applicable to use keyword arguments as follows:
```ruby
# Current CreatePalletFromCarton
def initialize(user, carton_id, carton_quantity, palletizing_bay_resource_id = nil, carton_palletizing = false, pallet_buildup = false)

# New version:
def initialize(user, carton_id, carton_quantity, palletizing_bay_resource_id: nil, carton_palletizing: false, pallet_buildup: false)


# Current NewPalletSequence:
def initialize(user_name, carton_id, pallet_id, carton_quantity, carton_palletizing = false, mix_rule_scope = nil, pallet_buildup = false)

# New version:
def initialize(user_name, carton_id, pallet_id, carton_quantity, carton_palletizing: false, mix_rule_scope: nil, pallet_buildup: false)
```

jcsjcs (changes requested):
Please attend to comments on your latest commit - to not provide arguments that are the same as the defaults.

PR #276, opened 2022-02-14
Orchard And Presorted Bins Putaway

lmatoti:
Summary of this change:
Display print_code instead of short_code


jcsjcs:
Re-implemented as a hotfix

PR #284, opened 2022-03-02
cons note number

miramanz:
Summary of this change:
cons note number

On all the lists/pallets and lists/pallet sequences grids wherever there are two columns called 'Cons Note Number', one should be removed- there must only be one
Then, the single 'Cons Note Number'  column should do a coalesce giving precedence to government_inspection_sheets.consignment_note_number (other being edi_in_consignment_note_number)
If any grid has a column called 'edi_consignment_note..'  it can remain there

### Changed
- . cons note number on all the lists/pallets and lists/pallet sequences grids

Lists | Pallets | 
Lists | Pallet Sequences |


db/migrations/202203011326_change_cons_note_number.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    sql = File.read(File.join(root_dir, 'vw_repacked_pallet_sequence_flat', '202203011326_vw_repacked_pallet_sequence_flat.sql'))
+    run 'DROP VIEW public.vw_repacked_pallet_sequence_flat;'
+    run sql
+  end
+
+  down do
+    root_dir = File.expand_path('../ddl/views', __dir__)
+
+    sql = File.read(File.join(root_dir, 'vw_pallet_sequence_flat', '000000000000_vw_pallet_sequence_flat.sql'))
```
jcsjcs, line 25: Please change all of these `000000000000` files and references to the last migration date for those views.
You can find those numbers from:
```
ag vw_pallet_sequence_flat db/migrations/ -l | sort | tail
```
and use the last one. e.g. if the last entry is `db/migrations/202201110951_add_legacy_attributes_to_views.rb`
Then rename:
`000000000000_vw_pallet_sequence_flat.sql` to `202201110951_vw_pallet_sequence_flat.sql`.

jcsjcs (changes requested):

PR #288, opened 2022-03-07
Presort Bin Staging

lmatoti:
Summary of this change:
Presort Bin Staging
1. Added  presort_staging_run.no_actual_ripeness_treatment
2. Staging execution: if run.no_actual_ripeness_treatment is true, only stage bin with actual_ripeness_treatment = null


lib/raw_materials/interactors/presort_staging_run_interactor.rb
```diff
@@ -8,6 +8,7 @@ def create_presort_staging_run(params) # rubocop:disable Metrics/AbcSize
       legacy = AppConst::CR_RMT.presort_legacy_data_fields.map { |f| [f, params[f]] }
       params[:legacy_data] = Hash[legacy] unless legacy.empty?
       res = validate_presort_staging_run_params(params)
+      params[:no_actual_ripeness_treatment] = false unless params[:actual_ripeness_treatment_id].nil_or_empty?
```
jcsjcs, line 11: You are setting a param value after the validation and then not referencing params again

jcsjcs (changes requested):

PR #305, opened 2022-03-16
Add new client rule: allocate_rebin_asset_numbers. 

miramanz:
Summary of this change:
Add new client rule: allocate_rebin_asset_numbers. If true allocate asset number from table bin_asset_numbers (order by last_used_at ASC, but null values MUST take precedence)

### Added
- .new client rule: allocate_rebin_asset_numbers
-  verify_rebin PDT and RMD apps

RMD | Raw Material | Rebin Verification


lib/client_rules/client_rmt_rules.rb
```diff
@@ -447,5 +456,11 @@ def enforce_mrl_check?(explain: false)
 
       setting(:enforce_mrl_check)
     end
+
+    def allocate_rebin_asset_numbers?(explain: false)
+      return 'Allocate asset number from bin_asset_numbers table.' if explain
```
jcsjcs, line 461: The explanation should include the fact that this is for rebins.

lib/messcada/services/rebin_verification.rb
```diff
@@ -32,6 +35,18 @@ def resolve_scanned_number_params
       ok_response
     end
 
+    def resolve_bin_asset_number
+      unless AppConst::CR_RMT.allocate_rebin_asset_numbers?
+        @bin_asset_number = carton_label_id.to_s
+        return ok_response
+      end
+
+      @bin_asset_number = RawMaterialsApp::RmtDeliveryRepo.new.get_available_bin_asset_numbers(1).map(&:last).first
```
jcsjcs, line 44: This can be `.first.last` instead of `.map(&:last).first`.

lib/messcada/services/rebin_verification.rb
```diff
@@ -32,6 +35,18 @@ def resolve_scanned_number_params
...
+      return failed_response("Couldn't find 1 available bin_asset_numbers in the system") unless bin_asset_number
```
jcsjcs, line 45: Just `Couldn't find available bin_asset_number in the system`.

jcsjcs (changes requested):

PR #306, opened 2022-03-17
 Ensure that the following client rules are used and not the .env.local values

miramanz:
Summary of this change:
Ensure that the following client rules are used and not the .env.local values:
DELIVERY_CAPTURE_CONTAINER_MATERIAL => AppConst::CR_RMT.capture_container_material?
DELIVERY_CAPTURE_CONTAINER_MATERIAL_OWNER => AppConst::CR_RMT.capture_container_material_owner?

Default Values: make true for all, set as false for Matrooze fontein and Habata.
Make the Qty Bins field show as a label in the case of a single bin edit screen
Validation error: rmt_delivery_bins

### Changed
- . to use AppConst::CR_RMT.capture_container_material? instead of DELIVERY_CAPTURE_CONTAINER_MATERIAL
- to use AppConst::CR_RMT.capture_container_material_owner? instead of DELIVERY_CAPTURE_CONTAINER_MATERIAL_OWNER

List | Bins | List | Recent
Raw Materials | Deliveries | Deliveries | edit


lib/production/ui_rules/rebin_rule.rb
```diff
@@ -12,8 +12,8 @@ def generate_rules
 
       @rules[:show_nett_weight] = AppConst::DELIVERY_CAPTURE_BIN_WEIGHT_AT_FRUIT_RECEPTION
       @rules[:capture_inner_bins] = AppConst::DELIVERY_CAPTURE_INNER_BINS && !@form_object.rmt_inner_container_type_id.nil?
-      # @rules[:capture_container_material] = AppConst::DELIVERY_CAPTURE_CONTAINER_MATERIAL
-      @rules[:capture_container_material_owner] = AppConst::DELIVERY_CAPTURE_CONTAINER_MATERIAL_OWNER
+      # @rules[:capture_container_material] = AppConst::CR_RMT.capture_container_material?
+      @rules[:capture_container_material_owner] = AppConst::CR_RMT.capture_container_material_owner?
```
jcsjcs, line 16: These rules are not necessary - just use the `AppConst` client rules.

lib/raw_materials/ui_rules/rmt_bin_rule.rb
```diff
@@ -76,37 +76,70 @@ def set_show_fields # rubocop:disable Metrics/AbcSize
...
       fields[:scrapped] = { renderer: :label, as_boolean: true }
     end
 
     def common_fields # rubocop:disable Metrics/AbcSize
-      fields = {
+      {
         bin_asset_number: { renderer: :label },
-        qty_bins: { required: true },
-        bin_fullness: { renderer: :select, options: AppConst::BIN_FULLNESS_OPTIONS, caption: 'Bin Fullness', required: true, prompt: true },
+        qty_bins: { renderer: :label },
```
jcsjcs, line 107: This should render with `include_hidden_field` so that the qty bins is never null.

jcsjcs (changes requested):

PR #307, opened 2022-03-21
FG Orders: added order value attrs

ViljoenFritz:
CHANGELOG entry:

### Added
- Order value (carton price)
- Order value (Kg price)
- Order item customer price history
### Changed
- set price per kg and price per carton always visible
### Fixed
- order item price history now filters even if values where null

look at orders >> order items


lib/finished_goods/interactors/order_interactor.rb
```diff
@@ -142,26 +141,38 @@ def col_defs(rpt, id, for_multiselect = false) # rubocop:disable Metrics/AbcSize
...
           act.separator
           act.popup_view_link '/development/statuses/list/order_items/$col1$',
                               col1: 'id', icon: 'information-solid', text: 'status', title: 'Status'
         end
+
         rpt.ordered_columns.each do |column|
           if %w[carton_quantity price_per_carton price_per_kg].include? column.name
-            hide = (column.name == 'price_per_kg' && !pricing_per_kg) || (column.name == 'price_per_carton' && pricing_per_kg)
-            mk.col(column.name, column.caption, { width: column.width, data_type: column.data_type, editable: true, cellEditor: 'numericCellEditor', hide: hide })
+            mk.col(column.name, column.caption, { width: column.width, data_type: column.data_type, editable: true, cellEditor: 'numericCellEditor' })
```
ViljoenFritz, line 175: can we cast this cell editor as currency

lib/finished_goods/entities/order.rb
```diff
@@ -40,6 +40,8 @@ class Order < Dry::Struct
     attribute :order_number, Types::String
     attribute :remarks, Types::String
     attribute :pricing_per_kg, Types::Bool
+    attribute :order_total_price_per_carton, Types::Decimal
+    attribute :order_total_price_per_kg, Types::Decimal
```
ViljoenFritz, line 44: can these types be changed to a currency attribute

lib/finished_goods/ui_rules/order_rule.rb
```diff
@@ -45,6 +45,12 @@ def set_show_fields # rubocop:disable Metrics/AbcSize
                                       with_value: !@form_object.pricing_per_kg,
                                       as_boolean: true }
       fields[:active] = { renderer: :label, as_boolean: true }
+      fields[:order_total_price_per_carton] = { renderer: :label,
+                                                caption: 'Order Total (Price per Carton)',
+                                                with_value: UtilityFunctions.delimited_number(@form_object.order_total_price_per_carton) }
```
ViljoenFritz, line 50: I dont like using the util_func here would rather have the dry struct handle it

lib/finished_goods/interactors/order_interactor.rb
jcsjcs: There is no currency editor. The grid would also need a column for the currency symbol.

lib/finished_goods/repositories/order_repo.rb
```diff
@@ -105,6 +107,9 @@ def find_order_item(id)
       hash[:loads] = select_values(:pallets, :load_id, id: pallet_ids).uniq
       hash[:load_id] = hash[:loads].first
       hash[:pricing_per_kg] = get(:orders, :pricing_per_kg, hash[:order_id])
+      hash[:carton_price_total] = select_values(:pallet_sequences, :carton_quantity, order_item_id: id).sum * hash[:price_per_carton].to_f
+      hash[:weight_price_total] = select_values(:pallet_sequences, :nett_weight, order_item_id: id).sum * hash[:price_per_kg].to_f
```
jcsjcs: Using floats is not very accurate - it might be better to use `BigDecimal` to get consistent results.
```ruby
 * (hash[:price_per_carton] || AppConst::BIG_ZERO)
```

lib/finished_goods/validations/order_schema.rb
```diff
@@ -54,6 +54,14 @@ class OrderItemContract < Dry::Validation::Contract
     rule(:actual_count_id) do
       key.failure('must be filled or size reference must be filled') if values[:actual_count_id].nil_or_empty? && values[:size_reference_id].nil_or_empty?
     end
+
+    rule(:price_per_carton) do
+      key.failure('must be filled') if values[:price_per_carton].nil_or_empty? && !BaseRepo.new.get(:orders, :pricing_per_kg, values[:order_id])
```
jcsjcs, line 59: Rather pass `price_per_kg` boolean as an option to the Contract.

jcsjcs (changes requested):

db/migrations/202203222143_alter_order_items_prices.rb
```diff
@@ -0,0 +1,35 @@
...
+
+
+    run 'ALTER TABLE order_items
+         ALTER COLUMN price_per_carton TYPE NUMERIC(16, 2),
+         ALTER COLUMN price_per_kg TYPE NUMERIC(16, 2);'
+
+    run File.read(File.join(root_dir, 'vw_count_order', '202203222143_vw_count_order.sql'))
+    run File.read(File.join(root_dir, 'vw_market_dashboard_data', '202203222143_vw_market_dashboard_data.sql'))
+    run File.read(File.join(root_dir, 'vw_market_dashboard_varieties', '202203222143_vw_market_dashboard_varieties.sql'))
+  end
```
jcsjcs, line 17: What are these view for?

If you create a view in a migration that does not already exist in the db:
* Do not keep the `000000000000` migration file.
* Do not include DROP in the `up` - there is no such view.
* The `down` should just be DROPs
* Change the ALTER to use the ruby migration DSL: `alter_table(:order_items)`...

jcsjcs (changes requested):

lib/finished_goods/entities/order.rb
jcsjcs: There is no currency type.

lib/finished_goods/repositories/order_repo.rb
```diff
@@ -57,10 +57,12 @@ def find_order(id) # rubocop:disable Metrics/AbcSize
       hash[:customer_payment_term_set] = MasterfilesApp::FinanceRepo.new.for_select_customer_payment_term_sets(
         where: { Sequel[:customer_payment_term_sets][:id] => hash[:customer_payment_term_set_id] }, active: nil
       ).flatten.first
-      Order.new(hash.merge(order_value(id)))
+      hash.merge!(order_item_value(order_id: id))
+
+      Order.new(hash)
```
jcsjcs, line 62: No need for the extra hash copy.
```ruby
Order.new(hash.merge(order_item_value(order_id: id)))
```

lib/finished_goods/repositories/order_repo.rb
```diff
@@ -107,8 +109,7 @@ def find_order_item(id)  # rubocop:disable Metrics/AbcSize
       hash[:loads] = select_values(:pallets, :load_id, id: pallet_ids).uniq
       hash[:load_id] = hash[:loads].first
       hash[:pricing_per_kg] = get(:orders, :pricing_per_kg, hash[:order_id])
-      hash[:carton_price_total] = select_values(:pallet_sequences, :carton_quantity, order_item_id: id).sum * hash[:price_per_carton].to_f
-      hash[:weight_price_total] = select_values(:pallet_sequences, :nett_weight, order_item_id: id).sum * hash[:price_per_kg].to_f
+      hash.merge!(order_item_value(order_item_id: id))
 
       OrderItem.new(hash)
```
jcsjcs, line 114: Don't do the extra hash copy when it is not necessary - include the merge here)

lib/finished_goods/ui_rules/order_rule.rb
jcsjcs: Dry struct is not for output formatting.

jcsjcs (changes requested):

lib/finished_goods/repositories/order_repo.rb
ViljoenFritz: cool

lib/finished_goods/interactors/order_interactor.rb
ViljoenFritz: Currently the  coded_grid  columns above does not respect the delimited_1000 setting in order_items.yml

PR #315, opened 2022-03-29
FG Allocations

ViljoenFritz:
Summary of this change:
Implement can_allocate_pre_stock_pallets? setting
Propagate inspection info to pallet if AppConst::CR_FG.use_continuous_govt_inspection_sheets?

CHANGELOG entry:

### Added
- client setting can_allocate_pre_stock_pallets
- govt_inspection_pallet tests
### Changed
- govt_inspection_pallets repos 
### Fixed
- pallet factory

- allocate non stock pallets to load
- on using kr client die govt inspections on pallets and test if inspection result is on pallet


lib/finished_goods/repositories/govt_inspection_repo.rb
```diff
@@ -104,30 +104,36 @@ def cancel_govt_inspection_sheet(id, user)
...
 
-      all_hash(:govt_inspection_pallets, govt_inspection_sheet_id: id).each do |govt_inspection_pallet|
+      govt_inspection_pallet_ids = select_values(:govt_inspection_pallets, :id, govt_inspection_sheet_id: id)
+      finish_govt_inspection_pallet(govt_inspection_pallet_ids, user)
+    end
+
+    def finish_govt_inspection_pallet(ids, user = nil)
+      all_hash(:govt_inspection_pallets, id: ids).each do |govt_inspection_pallet|
+        reinspection = get(:govt_inspection_sheets, :reinspection, govt_inspection_pallet[:govt_inspection_sheet_id])
+        status = reinspection ? 'MANUALLY REINSPECTED BY GOVT' : 'MANUALLY INSPECTED BY GOVT'
```
jcsjcs, line 122: Unless this is called from somewhere else, you can just pass in the status from `finish_govt_inspection_sheet`.

**Later:** I see it is indeed called from elsewhere. I think these methods can be simplified in another way, perhaps by splitting up some more and separating the loop from the pallet update and making the check of the inspection/re-inspection external...

Also you can pass the user to `create_govt_inspection_pallet`.

lib/finished_goods/test/factories/govt_inspection_factory.rb
```diff
@@ -48,7 +62,7 @@ def create_govt_inspection_sheet(opts = {})
     end
 
     def create_govt_inspection_pallet(opts = {})
-      pallet_id = create_pallet
+      pallet_id = create_pallet(force_create: true)
```
jcsjcs, line 65: The `force_create` should be handled in the calling code (if needed), and this line should be:
```ruby
opts[:pallet_id] ||= create_pallet
# ...
```
And the `default` does not include `:pallet_id`.

Run the scaffold to see this pattern.

jcsjcs (changes requested):

lib/finished_goods/repositories/govt_inspection_repo.rb
ViljoenFritz: If I had time I would refactor the interactor to merge the govt_inspection_sheet and pallet interactor. 
also I would then move the tripsheet stuff to its own interactor

lib/finished_goods/test/factories/govt_inspection_factory.rb
```diff
@@ -2,32 +2,46 @@
 
 module FinishedGoodsApp
   module GovtInspectionFactory
+    def self.included(klass)
+      klass.class_eval do
+        include MasterfilesApp::PartyFactory
+        include MasterfilesApp::DepotFactory
+        include MasterfilesApp::PackagingFactory
+        include MesscadaApp::PalletFactory
+        include MasterfilesApp::TargetMarketFactory
```
ViljoenFritz, line 11: Please check this pattern.
This will take some thinking to get every thing right. but the upside is, for the tests its then only one include. 

it saves a lot of time when setting up tests

lib/finished_goods/test/interactors/test_govt_inspection_pallet_interactor.rb
```diff
@@ -5,63 +5,102 @@
 module FinishedGoodsApp
   class TestGovtInspectionPalletInteractor < MiniTestWithHooks
     include GovtInspectionFactory
-    include MasterfilesApp::PartyFactory
-    include MasterfilesApp::DepotFactory
-    include MasterfilesApp::PackagingFactory
```
ViljoenFritz, line 10: see previous comment

lib/finished_goods/test/interactors/test_govt_inspection_pallet_interactor.rb
```diff
@@ -5,63 +5,102 @@
...
       res = interactor.create_govt_inspection_pallet(attrs)
       assert res.success, "#{res.message} : #{res.errors.inspect}"
-      assert_instance_of(GovtInspectionPallet, res.instance)
+      assert_instance_of(GovtInspectionPalletFlat, res.instance)
       assert res.instance.id.nonzero?
     end
 
+    def test_create_govt_inspection_pallet_auto_fills_passed
+      AppConst::CR_FG.stub(:use_continuous_govt_inspection_sheets?, true) do
+        assert AppConst::CR_FG.use_continuous_govt_inspection_sheets?
```
ViljoenFritz, line 30: Thanks for the assist, I really like this way of defining the tests

ViljoenFritz:
For the scaffold, consider to only create a factory for the current table being scaffolded.
As example, when there is a pallet_id on the table we get a infinity scroll where that section just balloons.
It also forces the dev to not duplicate the factories

lib/finished_goods/test/factories/govt_inspection_factory.rb
jcsjcs: It is _roughly_ a good pattern. It does need careful setup, though, and great care has to be taken to avoid circuar refs.
A general rule of thumb could be to never implement this in masterfile factories.

jcsjcs:
> 
Yes, but.

The problem is then there is no cover when introducing dependent tables with the current table.

PR #319, opened 2022-03-31
FG_Govt_Inspections

ViljoenFritz:
Summary of this change:

Propagate inspection info to pallet if AppConst::CR_FG.use_continuous_govt_inspection_sheets

CHANGELOG entry:

### Added
- .
### Changed
-  for Kr saves govt_inspection result to pallet when use_continuous_govt_inspection_sheets iss true
### Fixed
- .

add pallet to govt inspection sheet test if result value is on pallet table


lib/messcada/test/factories/pallet_factory.rb
```diff
@@ -44,7 +53,7 @@ def create_pallet(opts = {})
         reinspected: false,
         scrapped: false,
         carton_quantity: Faker::Number.number(digits: 4),
-        govt_inspection_passed: false,
+        govt_inspection_passed: nil,
```
jcsjcs, line 56: Why would this be `nil`? The default is `false`.

jcsjcs (changes requested):

lib/messcada/test/factories/pallet_factory.rb
ViljoenFritz: reverted

PR #323, opened 2022-04-02
No Merge - Change call back between route and interactor

ViljoenFritz:
No Merge

Proposed change on how the interactor replays to route. 

no longer a need for repo.find_tabel_name, as the query is already in the yml. 

no longer a need to maintain route - row_keys as it wall always match


jcsjcs:
Nice idea!
Simplifies the route code which is always a plus.
Probably the only concern could be performance (especially if there's something like a pallet view in the SQL), but probably not a lot slower than using a `Flat` entity for most cases...

Thanks, I'll think about this a bit more.

ViljoenFritz:
I believe it would in some cases be a performance boost, as Postgres can optimize the query for only the single id. In most cases using the Flat find methods, there are multiple DB hits, where the instance is assembled in ruby.

PR #324, opened 2022-04-03
Feature/mrl bulk refactored

WernerVdM97:
Summary of this change:
Implemented bulk capture chemical requirements spec. 

CHANGELOG entry:

### Added
- All necessary scaffolding for mrl_chemical_requirements
- Inline editable grid for bulk capture
- two views (edit, view) for the mrl_chemical_requirements grid
### Changed
- Removed redundant scaffolding
- reworked tests


lib/masterfiles/ui_rules/mrl_chemicals_requirement_rule.rb
```diff
@@ -0,0 +1,160 @@
...
+
+    def update_cultivar_group_columns # rubocop:disable Metrics/AbcSize
+      mrl_req_repo = MasterfilesApp::MrlRequirementRepo.new
+
+      actions = []
+      actions << OpenStruct.new(type: :change_select_value, dom_id: 'mrl_chemicals_requirement_cultivar_id', value: [])
+      actions << OpenStruct.new(type: :replace_input_value, dom_id: 'mrl_chemicals_requirement_qa_standards', value: [])
+      actions << OpenStruct.new(type: :replace_input_value, dom_id: 'mrl_chemicals_requirement_packed_tm_groups', value: [])
+      actions << OpenStruct.new(type: :replace_input_value, dom_id: 'mrl_chemicals_requirement_target_markets', value: [])
+      actions << OpenStruct.new(type: :replace_input_value, dom_id: 'mrl_chemicals_requirement_target_customers', value: [])
```
jcsjcs, line 135: Replace input value is for input fields, not multi or select. The value should be a string.
Here you should use `replace_multi_options` or `replace_select_options`.
`change_select_value` should also return a string value.

lib/masterfiles/ui_rules/mrl_chemicals_requirement_rule.rb
```diff
@@ -0,0 +1,160 @@
...
+
+      return json_actions(actions) if params[:changed_value] == '' || params[:changed_value].nil?
+
+      qa_std_options = mrl_req_repo.for_select_qa_stds_and_mrl_req_id where: { cultivar_group_id: params[:changed_value] }
+      tmg_options = mrl_req_repo.for_select_ptm_groups_and_mrl_req_id where: { cultivar_group_id: params[:changed_value] }
+      tm_options = mrl_req_repo.for_select_tms_and_mrl_req_id where: { cultivar_group_id: params[:changed_value] }
+      tc_options = mrl_req_repo.for_select_tcs_and_mrl_req_id where: { cultivar_group_id: params[:changed_value] }
+
+      actions << OpenStruct.new(type: :replace_multi_options,
```
jcsjcs, line 144: These actions should not include the blank actions from above.
A better way would be to set the local variable to either an empty array or values from the repo - depending on the changed value. Then `replace_multi_options` will be cleared or assigned as applicable (if that is indeed what you want).

routes/masterfiles/quality.rb
```diff
@@ -1552,5 +1552,55 @@ class Nspack < Roda
...
+    r.on 'bulk_capture' do
+      interactor = MasterfilesApp::MrlChemicalsRequirementInteractor.new(current_user, {}, { route_url: request.path, request_ip: request.ip }, {})
+      r.on 'change', String do |change_field|
+        handle_ui_change(:mrl_chemicals_requirement, change_field.to_sym, params, { field: change_field.to_sym })
+      end
+
+      r.is do
+        check_auth!('quality', 'read')
+        check_auth!('quality', 'new')
+        check_auth!('quality', 'edit')
```
jcsjcs, line 1567: You only need to check `edit` here, read is implied and _usually_ a user will not have `edit` without `new`.

jcsjcs (changes requested):

Gemfile
```diff
@@ -25,6 +26,7 @@ gem 'roda-symbolized_params'
 gem 'rodauth'
 gem 'rouge'
 gem 'rubocop'
+gem 'ruby-debug-ide'
```
jcsjcs, line 29: Developer gems must be within the `:development` block so that they do not get installed on live sites.

jcsjcs (changes requested):

PR #327, opened 2022-04-04
hotfix to update pallets.govt_first_inspected_at

ViljoenFritz:
Summary of this change:

Changed the inspection application not to update govt_first_inspected_at  for pallets
Added a migration to restore the original inspection date on pallets.


db/migrations/202204041842_fix_pallet_govt_first_inspected_date.rb
```diff
@@ -0,0 +1,33 @@
+Sequel.migration do
+  up do
```
jcsjcs, line 2: This must be a scriptfix, not a migration, please.

jcsjcs (changes requested):

db/migrations/202204041842_fix_pallet_govt_first_inspected_date.rb
ViljoenFritz: changed to script

PR #336, opened 2022-04-08
Locations

lmatoti:
Summary of this change:
Locations
- This sets locations.contains_row_positions=true for all locations of type BAY/DECK
and locations.is_row_position = true for all locations of type BAY_POSITION/DECK_POSITION


config/client_settings_loader.rb
```diff
@@ -110,6 +110,15 @@ class AppClientSettingsLoader
...
                                     desc: 'The code for location types that serve as bays in cold storage. Default is "DECK"' },
+    LOCATION_TYPES_COLD_BAY_DECK_POSITION: { env_key: 'LOCATION_TYPES_COLD_BAY_DECK_POSITION',
+                                             default: 'DECK_POSITION',
+                                             desc: 'The code for location types that serve as bays in cold storage. Default is "DECK"' },
+    LOCATION_TYPES_COLD_BAY_POSITION: { env_key: 'LOCATION_TYPES_COLD_BAY_POSITION',
+                                        default: 'BAY_POSITION',
+                                        desc: 'The code for location types that serve as bays in cold storage. Default is "DECK"' },
+    LOCATION_TYPES_COLD_BAY: { env_key: 'LOCATION_TYPES_COLD_BAY',
+                               default: 'BAY',
+                               desc: 'The code for location types that serve as bays in cold storage. Default is "DECK"' },
```
jcsjcs, line 121: These should not be client settings if they're just used for the script.

scriptfixes/set_row_position_fields_for_decs_and_bays.rb
```diff
@@ -0,0 +1,78 @@
...
+# -----------------------
+# New flags contains_row_positions and is_row_position have been implemented for the
+# location movement application. These flags need to set for the locations so they
+# can work for that application.
+
+class SetRowPositionFieldsForDecsAndBays < BaseScript
+  def run # rubocop:disable Metrics/AbcSize
+    deck_ids = DB[:locations]
+               .join(:location_types, id: :location_type_id)
+               .where(location_type_code: AppConst::LOCATION_TYPES_COLD_BAY_DECK)
```
jcsjcs, line 18: This should not use client settings, Just use args to the script.
```ruby
deck_type = args[0] || 'DECK'
deck_pos = args[1] || 'DECK_POSITION'
bay_pos = args[2] || 'BAY_POSITION'
```

jcsjcs (changes requested):

PR #338, opened 2022-04-11
Patch | Multiple load allocate

ViljoenFritz:
Summary of this change:

allow separating the load number from the pallet number using
semi colons, spaces and full stops

CHANGELOG entry:

### Added
- .
### Changed
- . tests
### Fixed
- .


lib/finished_goods/services/parse_load_string.rb
```diff
@@ -6,52 +6,55 @@ class ParseLoadString < BaseService
 
     def initialize(parse_string)
       @parse_string = parse_string
+      @pattern = /\n|,|\t| |\.|;|\r/
```
jcsjcs, line 9: I don't like the splitting on `.` and space. It could be confusing to the user. e.g. if there is an accidental space in the middle of a pallet number, it should be shown as a single pallet number validation error rather than two pallet number validation errors. Or in this case the user might get a validation message that the identifiers and pallets do not match and might struggle to see that the problem is a space inside a pallet number. And as for `.`, I don't think I've ever seen a csv delimited by that, so rather treat it as data.

jcsjcs, line 9: I don't like the splitting on `.` and space. It could be confusing to the user. e.g. if there is an accidental space in the middle of a pallet number, it should be shown as a single pallet number validation error rather than two pallet number validation errors. Or in this case the user might get a validation message that the identifiers and pallets do not match and might struggle to see that the problem is a space inside a pallet number. And as for `.`, I don't think I've ever seen a csv delimited by that, so rather treat it as data.

jcsjcs (changes requested):

lib/finished_goods/services/parse_load_string.rb
ViljoenFritz: Cool, only splitting the string on  \n  ,  \t  ;  \r

PR #347, opened 2022-04-13
Feature/location storage rules

miramanz:
Summary of this change:
- storage rule definition changes
- Units in location and loading states
- PDT App: Store bins against an assignment

### Added
- . -storage rule definition master files
-  Units in location and loading states
- PDT App: Store bins against an assignment

Raw Materials | Locations | Storage Definations | Bin Storage Rule Types | New
Raw Materials | Locations | Storage Definations | Bin Storage Rules | New
RMD | Raw Material | Bin Storage (Assignment)


lib/raw_materials/interactors/bin_storage_interactor.rb
```diff
@@ -0,0 +1,93 @@
...
+    rescue Crossbeams::InfoError => e
+      failed_response(e.message)
+    rescue StandardError => e
+      failed_response(e.message)
+    end
+
+    def validate_location(scanned_location, location_ids)
+      location_id = repo.get_id(:locations, location_long_code: scanned_location)
+      return validation_failed_response(messages: { location: ['Location does not exist'] }) if location_id.nil_or_empty?
+      return validation_failed_response(messages: { location: ['Location is not one of the suggested locations'] }) unless location_ids.include? location_id
```
jcsjcs, line 33: You can use `validation_failed_message_response` here instead of `validation_failed_response`, and `nil?` instead of `nil_or_empty?` (`get_id` returns an integer or nil, nothing else):
```ruby
return validation_failed_message_response(location: ['Location does not exist']) if location_id.nil?
```

jcsjcs, line 33: You can use `validation_failed_message_response` here instead of `validation_failed_response`, and `nil?` instead of `nil_or_empty?` (`get_id` returns an integer or nil, nothing else):
```ruby
return validation_failed_message_response(location: ['Location does not exist']) if location_id.nil?
```

lib/raw_materials/interactors/bin_storage_interactor.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    def del_repo
+      RawMaterialsApp::RmtDeliveryRepo.new
+    end
+
+    def validate_duplicate_bin_asset_number_scans(params, scans)
+      error = {}
+      duplicate_scans = scans.find_all { |v| scans.count(v) > 1 }
+      params.find_all { |_k, v| duplicate_scans.include?(v) }.each do |k, _v|
+        error.store(k, ["Bin #{params[k]} scanned more than once"])
```
jcsjcs, line 79: Isn't `#{params[k]}` the same as `#{_v}` at this point?

jcsjcs, line 79: Isn't `#{params[k]}` the same as `#{_v}` at this point?

jcsjcs (changes requested):

PR #386, opened 2022-05-13
Feature/mrl chemical results

WernerVdM97:
Added a per chemical per target market view showing pass/fail MRL chemical results. Furthermore, upon the addition of a pre harvest chemical to a post harvest MRL result, a duplicate pre harvest MRL result is created/updated containing the same pre harvest chemical value.

### Added
- Routing and view for new chemical results grid.
- Additional columns to capture MRL chemical results.
- Additional columns to MRL results per target market.
- Checks for the addition of a duplicate pre harvest MRL result.
- Repo functions to accomodate the duplication of an MRL result.


lib/quality/interactors/mrl_chemicals_result_interactor.rb
```diff
@@ -112,16 +119,59 @@ def col_defs_for_mrl_chem_results_grid
...
+      id = nil
+      mrl_repo.transaction do
+        id = mrl_repo.create_mrl_result(res)
+        log_status(:mrl_results, id, "DUPLICATE MRL CREATED: num ingredients: #{res[:num_active_ingredients]}")
+        log_transaction
+      end
+      id
+    rescue Sequel::UniqueConstraintViolation
+      validation_failed_response(OpenStruct.new(messages: { waybill_number: ['This mrl result already exists'] }))
+    rescue Crossbeams::InfoError => e
```
jcsjcs, line 144: If there is no code that this method calls that can raise a `Crossbeams::InfoError` then don't rescue it.

lib/quality/interactors/mrl_result_interactor.rb
```diff
@@ -193,6 +212,44 @@ def mrl_passed_for?(mrl_result_id, mrl_requirement_id) # rubocop:disable Metrics
       result
     end
 
+    def row_defs_for_mrl_chem_results_grid(mrl_requirement_ids_and_tm_names, mrl_result_id) # rubocop:disable Metrics/AbcSize,
+      chemical_samples = QualityApp::MrlChemicalsResultRepo.new.chemical_ids_and_samples_for(mrl_result_id).to_h
+      chemicals = chemical_req_repo.chemicals_with_max_eu_level
+      cultivar_id = repo.get(:mrl_results, :cultivar_id, mrl_result_id)
+      season_id = repo.get(:mrl_results, :season_id, mrl_result_id)
```
jcsjcs, line 219: This can be more efficient (one db query) as:
```ruby
cultivar_id, season_id = repo.get(:mrl_results, %i[cultivar_id season_id], mrl_result_id)
```

lib/quality/interactors/mrl_result_interactor.rb
```diff
@@ -193,6 +212,44 @@ def mrl_passed_for?(mrl_result_id, mrl_requirement_id) # rubocop:disable Metrics
...
+      unique_tms = requirements_repo.target_market_types_and_ids_for(mrl_requirement_ids_and_tm_names.keys)
+      rows = []
+
+      mrl_requirement_ids_and_tm_names.to_a.each do |mrl_req_id, tm_name|
```
jcsjcs, line 223: You don't need the `to_a` here. `Hash.each {|k,v| ...}` will be the same as `Array.each {|k,v| ...}` in this case.

jcsjcs (changes requested):

lib/quality/interactors/mrl_chemicals_result_interactor.rb
```diff
@@ -13,10 +13,16 @@ def handle_inline_mrl_chem_result(params)
...
-    def update_or_create_mrl_chemicals_result(params)
+    def update_or_create_mrl_chemicals_result(params, callback = false)  # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
       res = calculate_and_validate_mrl_chemicals_result_params(params)
       return validation_failed_response(res) if res.failure?
 
+      # update or create duplicate pre harvest result if current result is post harvest but contains pre harvest chemical
+      save_chemical_result_to_duplicate_pre_harvest_result(res.to_h) if MasterfilesApp::ChemicalRepo.new.pre_harvest_chemical?(res[:chemical_id]) && mrl_repo.post_harvest_result?(res[:mrl_result_id]) && !callback
+
+      # res is overwritten in case of duplicate callback. Therefore revalidate to reobtain original values.
+      res = calculate_and_validate_mrl_chemicals_result_params(params) unless callback
```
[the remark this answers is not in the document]
WernerVdM97: Stepping through the code I noticed the following:
- Once line 21 performs the callback to `update_or_create_mrl_chemicals_result`, `res` is undefined at first but assigned a new value which is returned from validating `params`. 
- The duplicate chemical result record is then created/updated accordingly.
- However once we step back into line 21 after having created/updated the duplicate record , the value of `res ` is no longer its original value. It has been overwritten by the second instance thereof i.e. that of the duplicate chemical result.
- Therefore the original chemical result is never update/created, but instead the duplicate is update/created twice.

By parsing a duplicate of `res`, such as `to_h` effectively does, does not prevent `res` from being overwritten. Therefore I opted to revalidate `params` as those values are still the original.
Alternatively I considered moving line 21 to 17, before validation overwrites res. This would prevent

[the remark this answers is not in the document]
WernerVdM97: Okay, thanks for the info regarding `to_a` I have now removed the redundant validation in https://github.com/NoSoft-SA/nspack/pull/386/commits/bdb51ad4682ff98a23d2d61ae9fa1798ce9879c9

PR #401, opened 2022-05-20
Hotfix/govt_inspection_sheets_can_user

ViljoenFritz:
CHANGELOG entry:
### Fixed
-


lib/finished_goods/task_permission_checks/govt_inspection_pallet.rb
```diff
@@ -36,7 +36,7 @@ def check_pallet!(check)
       end
 
       def create_check
-        can_override = !Crossbeams::Config::UserPermissions.can_user?(@user, :reworks, :can_change_govt_inspected_pallets) if @user&.permission_tree
+        can_override = !Crossbeams::Config::UserPermissions.can_user?(@user, :reworks, :can_change_govt_inspected_pallets) if @user
```
jcsjcs, line 39: Does this mean that the user can override only if they are **not** allowed to change inspected pallets in reworks?
Also, why is the existence of `permission_tree` no longer important?

lib/finished_goods/ui_rules/govt_inspection_sheet_rule.rb
```diff
@@ -298,7 +298,7 @@ def add_rules # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, M
       rules[:refresh_tripsheet] = (@form_object.tripsheet_created || @form_object.tripsheet_loaded) && !@form_object.tripsheet_offloaded && @repo.refresh_tripsheet?(@form_object.id)
       rules[:print_tripsheet] = (@form_object.tripsheet_created || @form_object.tripsheet_loaded) && !@form_object.tripsheet_offloaded
       rules[:tripsheet_complete] = @repo.refresh_to_complete_offload?(@form_object.id)
-      rules[:can_change_govt_inspected_pallets] = @form_object.inspected && Crossbeams::Config::UserPermissions.can_user?(@options[:user], :reworks, :can_change_govt_inspected_pallets)
+      rules[:can_change_govt_inspected_pallets] = @form_object.inspected & @options[:user] && Crossbeams::Config::UserPermissions.can_user?(@options[:user], :reworks, :can_change_govt_inspected_pallets)
```
jcsjcs, line 301: You're missing an ampersand.

jcsjcs (changes requested):

lib/finished_goods/ui_rules/govt_inspection_sheet_rule.rb
ViljoenFritz: I was going for true & nil = false
not true && nil = nil

jcsjcs: Ah, OK, I see - a falsy result would be fine, though.

lib/finished_goods/task_permission_checks/govt_inspection_pallet.rb
ViljoenFritz: I wasn't testing this correctly

PR #402, opened 2022-05-20
Rebins

lmatoti:
Summary of this change:
Rebins
- Added calculated_rebin_delivery_id and rebin_pick_date fields to table and view
- Populated the field calculated_rebin_delivery_id when rebins are created
- Added calculated_rebin_delivery_id and rebin_pick_date to the bin grids


lib/raw_materials/repositories/rmt_delivery_repo.rb
```diff
@@ -903,5 +903,14 @@ def average_sample_full_bin_weight_for(rmt_delivery_id)
...
     end
+
+    def runs_most_recent_tipped_bins_rmt_delivery_id(production_run_rebin_id)
+      DB[:rmt_bins]
+        .where(Sequel.lit('rmt_delivery_id is not null'))
+        .where(production_run_tipped_id: production_run_rebin_id)
+        .order_by(Sequel.desc(:bin_tipped_date_time))
+        .limit(1)
+        .get(:rmt_delivery_id)
+    end
```
jcsjcs, line 914: This can be written in Sequel without the lits:
```ruby
DB[:rmt_bins]
  .exclude(rmt_delivery_id: nil)
  .where(production_run_tipped_id: production_run_rebin_id)
  .reverse(:bin_tipped_date_time)
  .limit(1)
  .get(:id)
```

jcsjcs (changes requested):

PR #409, opened 2022-05-24
Hotfix/de-allocate

ViljoenFritz:
Summary of this change:

set shipped_at: nil, when de-allocating a pallet from a load

CHANGELOG entry:
### Fixed
- [hotfix: set shipped_at: nil, when de-allocating a pallet from a load](https://github.com/NoSoft-SA/nspack/pull/409/commits/61bbe45ab2f753805903be979ca06363a172701a)


lib/finished_goods/services/unship_load.rb
```diff
@@ -60,7 +60,7 @@ def unship_pallets # rubocop:disable Metrics/AbcSize
       location_id = repo.get_id(:locations, location_type_id: location_type_id)
       raise Crossbeams::InfoError, 'Site location not defined, unable to unship pallet' if location_id.nil?
 
-      attrs = { shipped: false, exit_ref: nil, in_stock: true, location_id: location_id }
+      attrs = { shipped: false, shipped_at: nil, exit_ref: nil, in_stock: true, location_id: location_id }
```
jcsjcs, line 63: This must not happen on unship. Please revert.

jcsjcs (changes requested):

lib/finished_goods/services/unship_load.rb
ViljoenFritz: thanks, laps of judgement

PR #410, opened 2022-05-24
hotfix: fix allocate_pallet_check

ViljoenFritz:
Summary of this change:

fixed allocation check where a shipped pallet could be allocated to a new load

CHANGELOG entry:
### Fixed
- 

use the text box on allocate load to try and add a shipped pallet. it should raise a warning.


lib/finished_goods/task_permission_checks/load.rb
```diff
@@ -8,6 +8,7 @@ def initialize(task, load_id = nil)
         @task = task
         @repo = LoadRepo.new
         @id = load_id
+        @pallet_ids = @repo.select_values(:pallets, :id, load_id: load_id)
```
jcsjcs, line 11: What if `load_id` is nil and you get the full set of non-stock and non-allocated pallets?

lib/messcada/task_permission_checks/pallet.rb
```diff
@@ -12,8 +12,7 @@ def initialize(tasks, args = {})
         @load_id = @args[:load_id]
         @order_id = @args[:order_id]
         @check_pallet_numbers = Array(@args[:pallet_number] || @args[:pallet_numbers])
-        load_pallet_ids = @repo.select_values(:pallets, :id, load_id: @load_id) if @load_id
-        @check_pallet_ids = Array(@args[:pallet_id] || @args[:pallet_ids] || load_pallet_ids)
+        @check_pallet_ids = Array(@args[:pallet_id] || @args[:pallet_ids])
```
jcsjcs, line 15: I may be dumb, but how is this different?
The check was called with a `load_id`, now with the load's pallet ids (which _should_ be the same set as on the old line 15)

jcsjcs (changes requested):

lib/finished_goods/task_permission_checks/load.rb
ViljoenFritz: TaskPermissionCheck::Load will return failed_response 'Record not found'

jcsjcs: Then don't read the pallets table - add a check for unless load_id is nil. At the moment it look either like an oversight or an obscure requirement and it is a db hit we do not need.

lib/messcada/task_permission_checks/pallet.rb
jcsjcs: Please explain with an example. I still can't see it.

jcsjcs (changes requested):

lib/messcada/task_permission_checks/pallet.rb
ViljoenFritz: What happened wasthat TaskPermissionCheck::Pallet(:allocate) was given a load_id and a pallet_number. 

Because load_id got higher priority the check was on the currently allocated pallets instead of the to be allocated pallet_number. So it passed the check.

I added a test to check it

lib/finished_goods/task_permission_checks/load.rb
ViljoenFritz: I moved it

PR #417, opened 2022-05-30
Rmd View Bin

lmatoti:
Summary of this change:
Rmd View Bin
- Container material owner lookup fixed


lib/raw_materials/interactors/rmt_bin_interactor.rb
```diff
@@ -913,7 +913,7 @@ def move_location_bin(bin_number, location_id) # rubocop:disable Metrics/AbcSize
     def rmt_bin_attrs_for_display(rmt_bin_id)
       bin = repo.find_rmt_bin_flat(rmt_bin_id).to_h
       bin[:bin_load_id] = repo.get(:bin_load_products, :bin_load_id, bin[:bin_load_product_id])
-      bin[:material_owner] = repo.container_material_owner_for(bin[:rmt_material_owner_party_role_id], bin[:rmt_container_material_type_id])
+      bin[:material_owner] = repo.container_material_owner_for(bin[:rmt_material_owner_party_role_id])
```
jcsjcs, line 916: Add a method to `MasterfilesApp::PartyRepo` and just call it instead of `container_material_owner_for`
```ruby
    def fn_party_role_name_with_role(id)
      DB.get(Sequel.function(:fn_party_role_name_with_role, id))
    end
```
Then this can be re-used by other parts of the app.

If you don't need to include the name of the role (presumably it is always "CONTAINER_MATERIAL_OWNER" or something similar) then just call `MasterfilesApp::PartyRepo#fn_party_role_name` instead.

jcsjcs (changes requested):

PR #427, opened 2022-06-06
New Pallet Sequence

lmatoti:
Summary of this change:
New Pallet Sequence
- If sequence.cultivar is null, then calculate it from product_setup


lib/messcada/services/new_pallet_sequence_object.rb
```diff
@@ -14,12 +14,14 @@ def initialize(user_name, carton_id, carton_quantity, carton_palletizing: false,
...
 
-    def call
+    def call # rubocop:disable Metrics/AbcSize
       @carton = repo.find_carton(carton_id)
       return failed_response('Carton not found.') if carton.nil?
 
       params = pallet_params
       params.merge!(carton_params)
+      product_setup_id = repo.get_value(:carton_labels, :product_setup_id, id: @carton[:carton_label_id])
+      params[:cultivar_id] ||= ProductionApp::ProductionRunRepo.new.resolve_setup_cultivar_id(product_setup_id) unless product_setup_id
```
jcsjcs, line 24: Shouldn't this be `if`, not `unless`?

lib/production/interactors/production_run_interactor.rb
```diff
@@ -129,6 +129,7 @@ def create_pallet_from_carton(carton_id) # rubocop:disable Metrics/AbcSize, Metr
       res = nil
       repo.transaction do
         res = MesscadaApp::CreatePalletFromCarton.call(@user, carton_id, cpp[:cartons_per_pallet])
+        ras.kl.aat
```
jcsjcs, line 132: Looks like a bit of Jamaica got left behind in the code!

jcsjcs (changes requested):

PR #430, opened 2022-06-07
MRL pallet check service

WernerVdM97:
Summary of this change:
Created a new service which takes in a pallet ID and for each pallet sequence performs an MRL check. The MRL check applies for both pre harvest and post harvest results. The MRL requirement is derived from cultivar plus either a target customer, target market or packed target market group for the pallet sequence. The MRL result (sample) is derived from the farm plus PUC plus season for pre harvest or from the production run for post harvest checks.

CHANGELOG entry:

### Added
- mrl_pallet_check.rb under quality/services
- functions to flatten a QA standard into the encapsulated packed target market groups or target markets.


lib/quality/services/mrl_pallet_check.rb
```diff
@@ -0,0 +1,99 @@
...
+      end
+    end
+
+    def mrl_passed_for?(mrl_result_id, mrl_requirement_id) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      # only interested in pre harvest chemicals in the case of a pre_harvest_result
+      where = {}
+      where[:orchard_chemical] = true if @repo.get(:mrl_results, :pre_harvest_result, mrl_result_id)
+      chemical_samples = @mrl_chem_result_repo.chemical_ids_and_samples_for(mrl_result_id, where: where)
+      chemical_requirements = @mrl_chem_requirement_repo.chemical_ids_and_max_levels_for(mrl_requirement_id, where: where)
+      mrl_req = @mrl_req_repo.mrl_requirement_constraints_for(mrl_requirement_id)
```
jcsjcs, line 82: You could probably pass the `mrl_req` into this method, instead of passing in the `mrl_requirement_id` and re-reading the table.

jcsjcs (changes requested):

lib/masterfiles/repositories/qa_standard_repo.rb
```diff
@@ -44,5 +44,38 @@ def allmarkets_standard_exists_for?(season_id, qa_standard_type_id, exclude: {})
     def season_id_of_newest_qa_std
       DB[:qa_standards].order_by(:created_at).last&.dig(:season_id)
     end
+
+    def find_qa_standard_with_tm(tm_id, season_id)  # rubocop:disable Metrics/AbcSize
+      # see if a qa std exists which contains/encapsulates this tm. Given season
+      type_id = get_id(:qa_standard_types, qa_standard_type_code: AppConst::QA_STANDARD_MRL)
+      id = DB[:qa_standards].where(applies_to_all_markets: true, season_id: season_id, qa_standard_type_id: type_id).get(:id)
```
jcsjcs, line 51: As a small optimization, you can create a method that memoizes the type id:
```ruby
def mrl_qa_standard_type_id
  @mrl_qa_standard_type_id ||= get_id(:qa_standard_types, qa_standard_type_code: AppConst::QA_STANDARD_MRL)
end
```
Then instead of querying the db and setting `type_id`, you just call the method `mrl_qa_standard_type_id` wherever you need it in the repo and for the lifetime of the repo instance, the database will be queried only one time to get the value.

lib/quality/services/mrl_pallet_check.rb
```diff
@@ -0,0 +1,108 @@
...
+        mrl_requirements << mrl_requirement
+
+        # no requirement = success
+        next if mrl_requirement.nil?
+
+        # get mrl sample results for pre-harvest and post-harvest, assumes pallet sequence contains all necessary columns
+        pre_harvest_result_id = @repo.find_latest_pre_harvest_result_for(pallet_sequence[:farm_id], pallet_sequence[:puc_id], pallet_sequence[:season_id], pallet_sequence[:cultivar_id])
+        post_harvest_result_id = @repo.find_post_harvest_result_for(pallet_sequence[:production_run_id])
+
+        # no result equal failure?
```
jcsjcs, line 31: You can change this back to "No result equal**s** failure?"

lib/quality/services/mrl_pallet_check.rb
```diff
@@ -0,0 +1,108 @@
...
+      @repo = MrlResultRepo.new
+      @qa_std_repo = MasterfilesApp::QaStandardRepo.new
+      @mrl_req_repo = MasterfilesApp::MrlRequirementRepo.new
+      @mrl_chem_result_repo = QualityApp::MrlChemicalsResultRepo.new
+      @mrl_chem_requirement_repo = QualityApp::MrlChemicalsRequirementRepo.new
+      @prod_run_repo = ProductionApp::ProductionRunRepo.new
+    end
+
+    def call # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      mrl_requirements = []
```
jcsjcs, line 18: Rather than storing an array of all `mrl_requirement` objects, just set a boolean:
```ruby
found_nil_requirement - false
@repo.all_hash...
   mrl_requirement = find_mrl_requirement_id_for(pallet_sequence)
  found_nil_requirement = true if mrl_requirement.nil?
  next if mrl_requirement.nil?
  # ...
end
return success_response('Missing MRL requirement(s)') if found_nil_requirement
```
You could instead use a count and return the number of missing requirements (if that would be a useful part of the response to the user)

jcsjcs (changes requested):

lib/masterfiles/ui_rules/mrl_requirement_rule.rb
```diff
@@ -132,6 +136,8 @@ def add_behaviours
     end
 
     def update_cultivars
+      return json_actions([]) if params[:changed_value] == ''
```
jcsjcs, line 139: Surely you want to replace the select options with `[]` for each of `cultivar_group_id` and `cultivar_id`?
What happens when you test and un-select a season when you already have cultivar + group selected?

lib/masterfiles/ui_rules/mrl_requirement_rule.rb
```diff
@@ -83,9 +83,13 @@ def common_fields # rubocop:disable Metrics/AbcSize
 
       if @mode == :new
         season_id = @repo.season_id_of_newest_mrl_requirement
-        fields[:season_id][:selected] = season_id
-        fields[:cultivar_group_id][:options] = @cultivar_repo.for_select_cultivar_groups_for(season_id)
-        fields[:cultivar_id][:options] = @cultivar_repo.for_select_cultivars_for(season_id)
+        if season_id.nil? # applicable when the MRL requirements table has not been populated yet.
+          fields[:season_id][:prompt] = true
```
jcsjcs, line 87: Shouldn't cultivar and group options be [] in this case? - to be populated when the season is chosen.
Also: should the season not be set in the `@form_object`?

jcsjcs (changes requested):

lib/masterfiles/ui_rules/mrl_requirement_rule.rb
WernerVdM97: Yes that is better. I have refactored it accordingly.

lib/masterfiles/ui_rules/mrl_requirement_rule.rb
WernerVdM97: Yes I missed this.

lib/masterfiles/ui_rules/mrl_requirement_rule.rb
```diff
@@ -115,7 +111,16 @@ def make_form_object
     end
 
     def make_new_form_object
-      @form_object = new_form_object_from_struct(MasterfilesApp::MrlRequirement)
+      @form_object = OpenStruct.new(season_id: @repo.season_id_of_newest_mrl_requirement,
```
jcsjcs, line 114: Use the `merge_hash` option - see http://localhost:9296/developer_documentation/ui_rules.adoc

jcsjcs (changes requested):

PR #435, opened 2022-06-08
Container Types

lmatoti:
Summary of this change:
Container Types
- Replaced 'asset number' with load_id
- Change the DM report


db/migrations/202206071535_add_load_id_to_container_events.rb
```diff
@@ -0,0 +1,15 @@
+Sequel.migration do
+  up do
```
jcsjcs, line 2: Both the up and down should include a 
```sql
TRUNCATE TABLE container_events
```
call to drop all rows before the DDL change. (the data in the table is useless)
Then the `load_id` can be set to `null: false`.

db/migrations/202206071535_add_load_id_to_container_events.rb
```diff
@@ -0,0 +1,15 @@
...
+    alter_table(:container_events) do
+      drop_column :asset_number
+      add_foreign_key :load_id, :loads, key: [:id]
```
jcsjcs, line 5: No need to specify the key if it is going to be `id`.

jcsjcs (changes requested):

db/migrations/202206071535_add_load_id_to_container_events.rb
```diff
@@ -1,12 +1,18 @@
 Sequel.migration do
   up do
+    run <<~SQL
+      TRUNCATE container_events CASCADE
```
jcsjcs, line 4: Why do you have cascade? Are there tables with foreign keys to this table?

jcsjcs (changes requested):

PR #437, opened 2022-06-09
support scrapped pallet on request validation

WernerVdM97:
Summary of this change:
Added functionality to both eCert preverify and Titan request validation to support scrapped pallets.

CHANGELOG entry:

### Fixed
- In the case of `pallet.scrapped == true`, the SQL JOIN between pallets and pallet_sequences changes from:
    - JOIN pallets ON pallets.id = pallet_sequences.pallet_id, to
    - JOIN pallets ON pallets.id = pallet_sequences.scrapped_from_pallet_id


lib/messcada/entities/dataset_pallet_sequence.rb
```diff
@@ -360,5 +372,338 @@ def sql_base
 
       SQL
     end
+
+    def sql_base_if_scrapped
```
jcsjcs, line 376: Rather than create two methods where the SQL could get out of sync over time, just change the `sql_base` method:
```ruby
def sql_base
  join = if scrapped
           'JOIN pallets ON pallets.id = pallet_sequences.scrapped_from_pallet_id'
         else
           'JOIN pallets ON pallets.id = pallet_sequences.pallet_id'
         end
  <<~SQL
    SELECT
    -- ...
    FROM pallet_sequences
    #{join}
    LEFT JOIN pallets repacked_from_pallets ON repacked_from_pallets.id = pallet_sequences.repacked_from_pallet_id
    -- ...
  SQL
```

jcsjcs (changes requested):

PR #443, opened 2022-06-14
Hotfix/mrl testing hotfixes

WernerVdM97:
Summary of this change:
Fixed failure on adding a pallet to a govt inspection sheet by removing a redundant MRL check containing a syntax error.
Also added the correct validation to clone MRL requirement.
Lastly, improved UI clarity on preharvest MRL results.


lib/masterfiles/validations/mrl_requirement_schema.rb
```diff
@@ -37,19 +37,30 @@ class MrlRequirementContract < Dry::Validation::Contract
...
+      required(:cultivar_group_id).maybe(:integer)
+      required(:cultivar_id).maybe(:integer)
+      required(:qa_standard_id).maybe(:integer)
+      required(:packed_tm_group_id).maybe(:integer)
+      required(:target_market_id).maybe(:integer)
+      required(:target_customer_id).maybe(:integer)
+      required(:percentage_of).filled(:decimal)
+    end
+
+    rule(:qa_standard_id) do
```
jcsjcs, line 57: The rule should include all values it refers to.
An alternative way of checking is:
```ruby
rule(:qa_standard_id, :packed_tm_group_id, :target_market_id], :target_customer_id) do
  cnt = [values[:qa_standard_id], values[:packed_tm_group_id], values[:target_market_id], values[:target_customer_id]].compact.length
  base.failure '...' unless cnt == 1
end
```

jcsjcs (changes requested):

lib/masterfiles/validations/mrl_requirement_schema.rb
WernerVdM97: I see, thanks.

This code was leftover from before I started with MRL, so I reused it. I corrected it in the original implementation thereof as well.

PR #446, opened 2022-06-15
Pallets/Bins Tripsheet Destination Location

lmatoti:
Summary of this change:
Pallets/Bins Tripsheet Destination Location
- Added tripsheet_destination_location_id to Pallets/Bins
- Created 2 triggers to update this field whenever a trisheep is loaded/unloaded/offloaded
- Added field to Pallets/Bins grids


db/migrations/202206151023_add_tripsheet_destination_location_to_vw_pallets.rb
```diff
@@ -0,0 +1,17 @@
+Sequel.migration do
```
jcsjcs, line 1: This migration file's contents can be combined with `db/migrations/202206141633_add_tripsheet_destination_location_to_vw_bins_flat.rb`

db/ddl/functions/fn_monitor_tripsheet_loaded_update_stock_items_tripsheet_destination_location/202206131617_fn_monitor_tripsheet_loaded_update_stock_items_tripsheet_destination_location.sql
```diff
@@ -0,0 +1,29 @@
+CREATE OR REPLACE FUNCTION public.fn_monitor_tripsheet_loaded_update_stock_items_tripsheet_destination_location()
```
jcsjcs, line 1: Both triggers can use the same function which is called if either `loaded_at` or `unloaded_at` changes.
Then your logic can be:
if NEW.loaded_at IS NOT NULL AND NEW.unloaded_at IS NULL
  destination = NEW.planned_location_to_id
elseif NEW.loaded_at IS NULL AND NEW.unloaded_at IS NOT NULL
  destination = NULL
else
  Don't make any changes.

jcsjcs (changes requested):

db/ddl/functions/fn_update_stock_items_tripsheet_destination_location/202206131617_fn_update_stock_items_tripsheet_destination_location.sql
```diff
@@ -0,0 +1,34 @@
...
+    IF (update_tripsheet_destination_location = 1) THEN
+        IF (stock_type_code = 'BIN') THEN
+            EXECUTE 'UPDATE rmt_bins SET tripsheet_destination_location_id=$2 from vehicle_job_units WHERE vehicle_job_units.stock_item_id=rmt_bins.id AND vehicle_job_id = $1;'
+            USING OLD.id, tripsheet_destination_location_id;
+        ELSIF (stock_type_code = 'PALLET') THEN
+            EXECUTE 'UPDATE pallets SET tripsheet_destination_location_id=$2 from vehicle_job_units WHERE vehicle_job_units.stock_item_id=pallets.id AND vehicle_job_id = $1;'
+            USING OLD.id, tripsheet_destination_location_id;
+        END IF;
+    END IF;
+RETURN OLD;
```
jcsjcs, line 28: The trigger should return `NEW`, not `OLD`.

db/migrations/202206131612_add_tripsheet_destination_location_id_to_bin_and_pallets.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    # Functions and triggers to update rmt_bins/pallets tripsheet_destination_location_id
+    root_dir = File.expand_path('../ddl/functions', __dir__)
+    sql = File.read(File.join(root_dir, 'fn_update_stock_items_tripsheet_destination_location', '202206131617_fn_update_stock_items_tripsheet_destination_location.sql'))
+    run sql
+
+    run <<~SQL
+      CREATE TRIGGER monitor_tripsheet_loaded_update_stock_items_tripsheet_destination_location
+      AFTER UPDATE OF loaded_at
```
jcsjcs, line 18: Just combine these into one trigger with:
```sql
AFTER UPDATE OF loaded_at, offloaded_at
```

db/migrations/202206171159_add_tripsheet_destination_location_to_vw_pallet_sequence_flat.rb
```diff
@@ -0,0 +1,17 @@
+Sequel.migration do
```
jcsjcs, line 1: Group these three related view migrations into one `db/migration` file - no need for 3 separate files.

jcsjcs (changes requested):

PR #447, opened 2022-06-15
add validation to govt inspection sheet

WernerVdM97:
As per request from Heila:

On deletion of a govt inspection pallet, if pallet is repacked, a fail validation is returned to prevent the user from deleting the pallet off of the sheet.

### Changed
- `TaskPermissionCheck::GovtInspectionPallet`, function `delete_check`


lib/finished_goods/task_permission_checks/govt_inspection_pallet.rb
```diff
@@ -3,7 +3,7 @@
 module FinishedGoodsApp
   module TaskPermissionCheck
     class GovtInspectionPallet < BaseService
-      attr_reader :task, :entity, :args, :repo
+      attr_reader :task, :id, :entity, :args, :repo
```
jcsjcs, line 6: There's no need for a reader for `id` - it is never referenced outside the initialize.
In fact the `@id = id` line is redundant.

lib/finished_goods/task_permission_checks/govt_inspection_pallet.rb
```diff
@@ -59,6 +59,7 @@ def edit_check
 
       def delete_check
         return failed_response 'Cant delete Govt Inspection Pallet record if reinspection exists.' if !reinspection? && inspection_count > 1
+        return failed_response 'Cant delete Govt Inspection Pallet if pallet has been repacked.' if repo.get(:pallets, :repacked, repo.get(:govt_inspection_pallets, :pallet_id, id))
```
jcsjcs, line 62: Please write `Can't` or `Cannot`, but not `Cant`.
Cant: hypocritical and sanctimonious talk, typically of a moral, religious, or political nature.

jcsjcs (changes requested):

lib/finished_goods/task_permission_checks/govt_inspection_pallet.rb
WernerVdM97: I am referencing id on line 62 at the end of the line to get the pallet id.

jcsjcs: OK. missed that.

jcsjcs (approved):

PR #450, opened 2022-06-21
Cancel tripsheet

lmatoti:
Summary of this change:
Cancel tripsheet
- Clear destination location on tripsheet cancel
- Restore previous location if cancelling an already loaded tripsheet
- Must be able to move tripsheet item if it's being moved to prev location


lib/finished_goods/repositories/govt_inspection_repo.rb
```diff
@@ -541,5 +541,26 @@ def active_tripsheet_pallets(pallet_ids)
       SQL
       DB[qry, pallet_ids].select_map
     end
+
+    def move_tripsheet_items_back_to_previous_location(vehicle_job_id)
+      stock_type_code = DB[:stock_types]
+                        .join(:vehicle_jobs, stock_type_id: :id)
+                        .where(Sequel[:vehicle_jobs][:id] => vehicle_job_id)
+                        .get(:stock_type_code)
+      stock_items = select_values(:vehicle_job_units, :stock_item_id, vehicle_job_id: vehicle_job_id)
```
jcsjcs, line 550: If you change this query to join to pallets/bins as applicable and return the `previous_location_id` along with the `stock_item_id` you will not need to get the `previous_location_id` over and over in the loop below.

lib/finished_goods/repositories/govt_inspection_repo.rb
```diff
@@ -541,5 +541,26 @@ def active_tripsheet_pallets(pallet_ids)
...
+      stock_items.each do |stock_item_id|
+        if stock_type_code == AppConst::PALLET_STOCK_TYPE
+          previous_location_id = get_value(:pallets, :previous_location_id, id: stock_item_id)
+          business_process = AppConst::OPEN_TRIPSHEET_MOVE_PALLET_BUSINESS_PROCESS
+        elsif stock_type_code == AppConst::BIN_STOCK_TYPE
+          previous_location_id = get_value(:rmt_bins, :previous_location_id, id: stock_item_id)
+          business_process = AppConst::OPEN_TRIPSHEET_MOVE_BIN_BUSINESS_PROCESS
+        end
+
+        res = FinishedGoodsApp::MoveStock.call(stock_type_code, stock_item_id, previous_location_id, business_process, nil)
```
jcsjcs, line 560: There is no need for the `nil` at the end of the call - it is the default.

lib/finished_goods/repositories/govt_inspection_repo.rb
```diff
@@ -541,5 +541,26 @@ def active_tripsheet_pallets(pallet_ids)
...
+        raise res.message unless res.success
```
jcsjcs, line 561: Why are you raising a `RuntimeError` here? You should raise a`Crossbeams::InfoError`.

jcsjcs (changes requested):

PR #456, opened 2022-06-27
Add data migration script for multi loads

ViljoenFritz:
Summary of this change:

Add data migration script for MC to group loads as multi loads

NB: first migrate to latest. 
Only valid for shipped loads


scripts/data_migration_multi_loads.rb
```diff
@@ -0,0 +1,615 @@
...
+    raise Crossbeams::InfoError, "Load: #{errors.join(', ')} not shipped." unless errors.empty?
+  end
+
+  def delete_load_vehicle(load_id)
+    load_vehicle_id = DB[:loads].where(id: load_id).get(:load_vehicle_id)
+    load_container_id = DB[:load_vehicles].where(id: load_vehicle_id).get(:load_container_id)
+
+    DB[:loads].where(id: load_id).update(load_vehicle_id: nil)
+    DB[:load_vehicles].where(id: load_vehicle_id).delete
+    DB[:load_containers].where(id: load_container_id).delete
```
jcsjcs, line 90: Delete the container before the vehicle - otherwise you might get a foreign key violation.

jcsjcs (changes requested):

PR #457, opened 2022-06-27
Feature/temp tail

ViljoenFritz:
CHANGELOG entry:

### Added
- [Add AppConst::CR_FG.temp_tail_required_to_ship](https://github.com/NoSoft-SA/nspack/pull/457/commits/cbe66f40f3ee4bf7ab663314d880dc3d5bc868e1)
### Changed
- [Set requires_temp_tail to disabled when AppConst::CR_FG.temp_tail_required_to_ship](https://github.com/NoSoft-SA/nspack/pull/457/commits/d7f9576c994ff9514f7eb64bd94a8aa9e32d7a77) 

Loads: create or edit load where client setting is set. 
temptail required should be greyed out


lib/client_rules/client_fg_rules.rb
```diff
@@ -137,6 +141,7 @@ class ClientFgRules < BaseClientRules
             set_local_stock_via_pi: false,
             wo_fulfillment_pallet_warning_level: 2,
             enable_depot_returns: false,
+            temp_tail_required_to_ship: false,
```
jcsjcs, line 144: True for UD

lib/client_rules/client_fg_rules.rb
```diff
@@ -451,5 +460,11 @@ def default_phyto_inspection_depot(explain: false)
 
       setting(:default_phyto_inspection_depot)
     end
+
+    def temp_tail_required_to_ship(explain: false)
+      return 'Makes temp tail required on all loads' if explain
```
jcsjcs, line 465: Change the description - 'Makes temp tail required on all loads before they can be shipped'.

jcsjcs (changes requested):

PR #517, opened 2022-08-11
feature/container_code_pattern

FritzViljoen:
CHANGELOG entry:
load container_code for client UD must now conform to pattern 'abc1234567'

### Added
- regex test to truck arrival contract

on load do a truck arrival and update a container code


lib/finished_goods/validations/load_truck_arrival_contract.rb
```diff
@@ -13,7 +13,7 @@ class TruckArrivalContract < Dry::Validation::Contract
       required(:driver_cell_number).filled(Types::StrippedString)
       optional(:container).filled(:bool)
       # Container
-      optional(:container_code).maybe(Types::StrippedString, max_size?: 13)
+      optional(:container_code).maybe(Types::StrippedString, format?: AppConst::CR_FG.load_container_code_regex)
```
jcsjcs, line 16: If you implement the format check as a rule, you'll be able to test with different client codes.

Note: you should probably also use the `:pattern` attribute of the renderer to get the browser to do the check in the form itself. Quicker response to the user and better UX.

jcsjcs (changes requested):

lib/client_rules/client_fg_rules.rb
```diff
@@ -141,6 +145,7 @@ class ClientFgRules < BaseClientRules
             set_local_stock_via_pi: false,
             wo_fulfillment_pallet_warning_level: 2,
             enable_depot_returns: false,
+            load_container_code_regex: /^[a-zA-Z]{3}\d{7}$/,
```
jcsjcs, line 148: The regex requirement was incorrectly stated. Please change to 4 characters and 7 digits.

jcsjcs (changes requested):

lib/client_rules/client_fg_rules.rb
FritzViljoen: changed to /^[a-zA-Z]{4}\d{7}$/ for UD

lib/finished_goods/validations/load_truck_arrival_contract.rb
FritzViljoen: cool idea, I implemented the rule change

PR #583, opened 2022-10-14
Stock Movements

lmatoti:
Summary of this change:
Stock Movements
- Prevent stock_movements in/out of sealed location


lib/masterfiles/repositories/location_repo.rb
```diff
@@ -860,6 +860,29 @@ def storage_rule_type_location_assignments(location_assignment_ids, storage_rule
...
+    end
+
+    def location_current_event_status(id)
+      hash = DB[:location_coldroom_events]
+             .where(location_id: id)
+             .order_by(Sequel.desc(:created_at))
+             .first
+      return nil if hash.nil?
+
+      hash[:event_name]
```
jcsjcs, line 883: This looks for the current event on the location, not the CA_TREATMENT location above it.
Change `location_has_ca_treatment_assigment?` to a method that returns the id and use that to check.

In fact you could perhaps change your whole approach to use a block - something like this:
```ruby
# in the service:
check_ca_treatment_location(location_from_id) do |ca_treatment_location_id|
  status = locn_repo.location_current_event_status(ca_treatment_location_id)
  return failed_response("Stock item cannot be moved from a sealed location (status is #{status})") if AppConst::LOCATION_SEALED_CA_STATUSES.include?(status)
end

# In the repo:
def check_ca_treatment_location(location_id)
  ancestor = DB... # (Sequel as in previous comment)
  yield ancestor unless ancestor.nil?
end
```
Then the block will only be called (with the relevant location `id`) if there is an ancestor CA_TREATMENT location.

grid_definitions/dataminer_queries/ca_bin_locations.yml
```diff
@@ -1,30 +1,50 @@
...
+    "locations"."consumption_area",
+    "locations"."created_at",
+    "locations"."updated_at",
+    "location_types"."location_type_code",
+    "location_assignments"."assignment_code",
+    "location_storage_types"."storage_type_code",
+    "locations"."location_storage_definition_id",
+    "location_storage_definitions"."storage_definition_code",
+  	CASE 
+  		WHEN fn_current_status('locations', locations.id) IN ('GAS','SEALED_CA','SEALED_CA_SMARTFRESH','SEALED_RA_SMARTFRESH') THEN 'error'
```
jcsjcs, line 28: To be consistently correct, this should be checking against `location_coldroom_events`, not `status` (which could in future be something unrelated).

In the join clause, join to the latest coldroom event:
```sql
JOIN (SELECT DISTINCT ON (location_id) location_id, event_name
      FROM location_coldroom_events
      ORDER BY location_id, created_at DESC) AS cold_events ON cold_events.location_id = locations.id
```
And in the `WHEN`, check `cold_events.event_name` instead of `fn_current_status`.

jcsjcs (changes requested):

lib/masterfiles/repositories/location_repo.rb
```diff
@@ -860,6 +860,25 @@ def storage_rule_type_location_assignments(location_assignment_ids, storage_rule
...
+                 .where(descendant_location_id: id, assignment_code: AppConst::LOCATION_ASSIGNMENT_CA_TREATMENT)
+                 .get(:ancestor_location_id)
+      yield ancestor unless ancestor.nil?
+    end
+
+    def location_current_event_status(id)
+      hash = DB[:location_coldroom_events]
+             .where(location_id: id)
+             .order_by(Sequel.desc(:created_at))
+             .first
```
jcsjcs, line 876: Change `first` to `get(:event_name)` - you don't need to retrieve the whole record into a hash.
In fact you don't need to assign to anything because `get` will return the value or nil which is what you want to return.

jcsjcs (changes requested):

PR #586, opened 2022-10-19
standard pack sync

miramanz:
Summary of this change:
Change behavior of standard pack to populate pallet bases and stack types as well
If standard_pack.bin is true then user must only be able to select pallet_formats.pallet_base_ids and pallet_formats.pallet_stack_types_ids pallet_formats.bin is true OR vice versa
BinLocationRule - match on rule id rather than rule type id

### Changed
- . behavior of standard pack to populate pallet bases and stack types as well

### Fixed
- . BinLocationRule - to match on rule id rather than rule type id

*product setups
Production | Product Setups | Product Setup Templates | Product Setups | edit

Single Pallet Edit*
Reworks | Production | Pallets | Single Pallet Edit | New
Reworks | Finished Goods | Single Pallet Edit | New

Batch Pallet Edit*
Reworks | Production | Pallets | Batch Pallet Edit | New
Reworks | Finished Goods | Batch Pallet Edit | New


lib/production/interactors/product_setup_interactor.rb
```diff
@@ -110,6 +110,16 @@ def for_select_pallet_formats(pallet_base_id, pallet_stack_type_id, standard_pac
       )
     end
 
+    def for_select_pallet_stack_types(bin)
```
jcsjcs, line 113: This interactor method and the matching one in reworks run interactor use the same pattern.
Perhaps you can push the pattern down to PackagingRepo e.g. `for_select_stack_types_based_on_format_type(bin)` and call that from both interactors.

You can do the same for `for_select_bases_based_on_format_type(bin)`.

Also the next method - for cpp...

And you should be able to use these methods from `lib/production/ui_rules/packing_specification_wizard_packaging_rule.rb` too.

jcsjcs (changes requested):

PR #602, opened 2022-11-03
Add 'diff xml config' option to system resources

WernerVdM97:
Summary of this change:
Adds the option to compare a NsPack-generated config.xml file with the actual config.xml file on the NTD.
Access via the action menu on the system resources grid.

CHANGELOG entry:

### Added
- `MesserverRepo` function `fetch_config_xml_for(system_resource_id)` to get the config.xml file via an HTTP call to the NTD. 
- Routing, view and UI rule values to diff the fetched XML with a generated one.


lib/production/repositories/resource_repo.rb
```diff
@@ -781,6 +781,20 @@ def available_for_allocation_plant_resource_ids_for(where: {}, exclude: {}) # ru
...
+      ip = ntd[:ip_address]
+      port = ntd[:port]
+      raise Crossbeams::InfoError, "NTD IP:Port of NTD #{ntd[:system_resource_code]} not set." if ip.nil? || port.nil?
+
+      url = "http://#{ip}:#{port}/?Type=GetConfigFile"
+      http = Crossbeams::HTTPCalls.new
+      res = http.request_get(url)
+      raise Crossbeams::InfoError, "Failed to retreive NTD config.xml with IP: #{ip}" unless res.success
+
+      res.instance.body
```
jcsjcs, line 795: This should be part of the MesServer repo (all MesServer calls should be in the one repo).
So in the UI rule, do something like this:
```ruby
ip, port = repo.get(:system_resources, %i[ip_address port], id)
res = messerver_repo.module_config_xml(ip, port)
```
And return and handle failed responses instead of raising exceptions.

jcsjcs (changes requested):

lib/messerver/repositories/messerver_repo.rb
```diff
@@ -4,6 +4,18 @@ module MesserverApp
   class MesserverRepo
     include Crossbeams::Responses
 
+    def fetch_config_xml_for(system_resource_id) # rubocop:disable Metrics/AbcSize
+      ntd = DB[:system_resources].where(id: system_resource_id).first
```
jcsjcs, line 8: Rather:
```ruby
ip, code = get(:system_resources, %i[ip_address system_resource_code], system_resource_id)
```

lib/production/ui_rules/system_resource_rule.rb
```diff
@@ -65,6 +71,15 @@ def set_deploy_fields
                                with_value: RESOURCES::MODULE_DISTRO_TYPES.rassoc((@form_object.extended_config || {})['distro_type']).first }
     end
 
+    def config_diff_field
+      @form_object[:config_res] = MesserverApp::MesserverRepo.new.fetch_config_xml_for(@form_object[:id])
```
jcsjcs, line 75: This does not need to be part of the `@form_object` (nor does the generated config).
And add the error to `rules` to check in the view if it is not null.
```ruby
res = MesserverApp::MesserverRepo.new.fetch_config_xml_for(@form_object[:id])
if res.success
      { xml_diff: { left_caption: 'Generated Config',
                    right_caption: 'Existing Config',
                    left: @options[:generated_config],
                    right: res.instance,
                    max_pane_width: 900 } }
else
  rules[:fetch_error] = res.message unless res.success
end
```

lib/production/views/system_resource/diff_xml.rb
```diff
@@ -0,0 +1,26 @@
...
+      class DiffXml
+        def self.call(id, generated_config)
+          ui_rule = UiRules::Compiler.new(:system_resource, :diff_config, id: id, generated_config: generated_config)
+          rules   = ui_rule.compile
+
+          Crossbeams::Layout::Page.build(rules) do |page|
+            page.section do |section|
+              section.add_control control_type: :link, text: 'Back', url: '/list/system_resources', style: :back_button
+              if ui_rule.form_object.config_res.success
+                section.add_control control_type: :link, behaviour: :popup, text: 'Deploy', url: 'deploy_config', style: :button, icon: :download
```
jcsjcs, line 15: The url is not correct.

jcsjcs (changes requested):

PR #608, opened 2022-11-09
Feature/po in pallet gross weight change

WernerVdM97:
Summary of this change:
Removed previous handling of `pallet_gross_weight` field in `PO_in` service.
~Replaced with the following: if `pallet_gross_weight` is empty, `nett_weight_externally_calculated` is set to true and the `mass` field in the PO field is saved in `nett_weight` for each sequence.~
Spec changed: Now always saves `nett_weight` and sets `nett_weight_externally_calculated` true.

CHANGELOG entry:

### Added
- `Mass` column to manual intake inline edit grid.
- Validation to check if `mass` does not equal zero.
### Removed
- business logic around the `derive_nett_weight?` client rule to handle the saving of `pallet_gross_weight` for PO in files.

Receive two PO in files, one with `pallet_gross_weight` specified and one `nil`. 
Receive two PO files, one with `mass` specified and one equal to zero.
See EDI | Receipts | Recent for the results.


lib/edi/services/po_in.rb
```diff
@@ -21,6 +21,8 @@ def call # rubocop:disable Metrics/AbcSize
         @tot_cartons = sum_cartons(sequences)
         build_pallet(pallet_number, sequences.first)
         sequences.each do |sequence|
+          return failed_response("PO field 'Mass' cannot be 0.") if sequence[:mass].to_f.zero?
```
jcsjcs, line 24: This message should include some context - pallet number and sequence number - so that the user can find and rectify.

lib/edi/services/po_in.rb
```diff
@@ -171,7 +173,7 @@ def build_pallet(pallet_number, seq) # rubocop:disable Metrics/AbcSize, Metrics/
       orig_inspec_date = time_from_date_val(seq, :orig_inspec_date)
       inspec_date = time_from_date_val(seq, :inspec_date)
       intake_date = time_from_date_val(seq, :intake_date)
-      weighed_date = time_from_date_and_time(seq[:weighing_date], seq[:weighing_time])
+      # weighed_date = time_from_date_and_time(seq[:weighing_date], seq[:weighing_time])
```
jcsjcs, line 176: Why are we not storing the `gross_weight_measured_at` value any more?
This should still be applied if there is a gross weight.

lib/edi/services/po_in.rb
```diff
@@ -327,6 +326,7 @@ def build_sequence(pallet_number, seq) # rubocop:disable Metrics/AbcSize, Metric
         sell_by_code: seq[:sellbycode],
         product_chars: seq[:prod_char]   # ???
       }
+      rec[:record][:nett_weight] = seq[:mas] if parent[:record][:nett_weight_externally_calculated]
```
jcsjcs, line 329: Shouldn't that be `mass`, not `mas`?

jcsjcs (changes requested):

lib/edi/services/po_in.rb
```diff
@@ -228,6 +222,11 @@ def build_pallet(pallet_number, seq) # rubocop:disable Metrics/AbcSize, Metrics/
         temp_tail: seq[:temp_device_id],
         edi_in_inspection_point: seq[:inspect_pnt]
       }
+      if seq[:pallet_gross_mass].nil? || seq[:pallet_gross_mass].to_f.zero?
+        rec[:record][:nett_weight_externally_calculated] = true
```
jcsjcs, line 226: `nett_weight_externally_calculated` must be true for all pallets (with or without a gross).

lib/edi/services/po_in.rb
jcsjcs, line 329: `nett_weight` must be set for all pallet sequences - regardless of the pallet's `gross_weight`.

jcsjcs (changes requested):

lib/edi/services/po_in.rb
WernerVdM97: I implemented it in this way because in the spec Hans sent me he states: "If gross_weight is null:  - update pallet.nett_weight_externally_calculated to true". Maybe I interpreted this wrong or it is simply a typo?

Furthermore, if gross_weight is null, then I believe there is some trigger in the database that calculates its value and therefore `nett_weight_externally_calculated` should remain false. Correct me if I am wrong. 

However, what I realise in hindsight here is that I should probably add this true flag to each pallet sequence as well?

lib/edi/services/po_in.rb
WernerVdM97: Yes, I somehow missed this during testing...

lib/edi/services/po_in.rb
WernerVdM97: My mistake.

lib/edi/services/po_in.rb
WernerVdM97: Once again the spec Hans sent me seems to specify to specifically save `mass` to `nett_weight` only in the case of `gross_weight` equaling null. I will confirm with him.

jcsjcs: I have discussed this with him.

lib/edi/services/po_in.rb
WernerVdM97: Ah okay, I have cleared up with Hans about the spec changing.

I am setting `nett_weight_externally_calculated` to always be true.

jcsjcs: Gross is never calculated, it is always there as a consequence of weighing.

All the sequences get their nett weight from the mass value and this must remain a fact, so the pallet must have `nett_weight_externally_calculated` always set to true so that the sequences' nett is never re-calculated in our system from the gross (when provided) as we could otherwise end up with different nett values than were provided to us from the PO file.

lib/edi/services/po_in.rb
```diff
@@ -324,9 +323,11 @@ def build_sequence(pallet_number, seq) # rubocop:disable Metrics/AbcSize, Metric
         carton_quantity: seq[:ctn_qty].to_i,
         pick_ref: seq[:pick_ref],
         sell_by_code: seq[:sellbycode],
+        nett_weight_externally_calculated: true,
```
jcsjcs, line 326: There is no `nett_weight_externally_calculated` column on pallet sequences.

jcsjcs (changes requested):

lib/edi/services/po_in.rb
WernerVdM97: Okay, thanks for that clearing up.

lib/edi/services/po_in.rb
WernerVdM97: Noted.

PR #610, opened 2022-11-09
Reworks

miramanz:
Summary of this change:
- single pallet edit/sequence: create grid menu action on sequence grid to 'scrap cartons'- show multi-select grid to allow users to select cartons to be scrapped
- Also: the validation message appearing on 'scrap sequence' saying 'sequence cannot be scrapped should be in red (not green)
- Reworks: scrap cartons: Text area should say: 'Paste carton label ids' not 'carton ids'

### Added
- .  scrap cartons grid menu action to allow multi-select cartons to be scrapped

### Changed
- . the validation error message to red appearing on scrap sequence
- scrap cartons: text area should say: 'Paste carton labels' not 'Paste cartons'

Single Pallet Edit
Reworks | Production | Pallets | Single Pallet Edit | New | scrap cartons
Reworks | Finished Goods | Single Pallet Edit | New | scrap cartons

Scrap Cartons
Reworks | Production | Cartons | Scrap Carton| New


lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -1021,6 +1021,35 @@ def print_reworks_carton_label_for_sequence(sequence_id, params) # rubocop:disab
...
+      reworks_run_attrs = { user: @user.user_name,
+                            reworks_run_type_id: reworks_run_type_id,
+                            pallets_selected: "{ #{pallet_number} }",
+                            pallets_affected: "{ #{pallet_number} }",
+                            cartons_scrapped: repo.array_for_db_col(multiselect_list),
+                            changes_made: resolve_changes_made(before: before_attrs.sort.to_h,
+                                                               after: after_attrs.sort.to_h) }
+      repo.transaction do
+        carton_quantity = (repo.pallet_sequence_carton_quantity(sequence_id) - multiselect_list.count)
+        raise Crossbeams::TransactionFailError, 'Carton cannot be scrapped. Scrap the pallet sequence instead' unless carton_quantity > 1
```
jcsjcs, line 1039: Should that be `> 1` or `> 0` (`.positive?`)?
i.e. you are not allowing a sequence to exist if it has only one carton?

lib/production/interactors/reworks_run_interactor.rb
```diff
@@ -1021,6 +1021,35 @@ def print_reworks_carton_label_for_sequence(sequence_id, params) # rubocop:disab
...
+
+        repo.update(:cartons, multiselect_list, after_attrs)
+        repo.update(:pallet_sequences, sequence_id, carton_quantity: carton_quantity)
+        reworks_run_id = repo.create_reworks_run(reworks_run_attrs.merge(flag_attrs))
+        log_multiple_statuses(:cartons, multiselect_list, AppConst::REWORKS_ACTION_SCRAP_CARTON)
+        log_status(:reworks_runs, reworks_run_id, 'CREATED')
+        log_transaction
+      end
+      success_response('Cartons scrapped successfully', pallet_number: pallet_number)
+    rescue Crossbeams::InfoError => e
```
jcsjcs, line 1049: Shouldn't this be rescuing a `TransactionFailError`?

jcsjcs (changes requested):

lib/production/interactors/reworks_run_interactor.rb
miramanz: > 1 I'm not allowing users to scrap all cartons on a sequence (should have at least 1 carton left on a sequence). If that is the case the user should scrap the sequence instead.

jcsjcs: OK but then shouldn't your check be `unless carton_quantity.positive?`
i.e. you block if you will end up with 0 cartons on the pallet?, but you don't block if you end up with 1 carton on the sequence?
```ruby
seq_ctns = 20
list_count = 19
ctn_q = seq_ctns - list_count
#=> 1
puts 'ERR' unless ctn_q > 1
#=> ERR

puts 'ERR' unless ctn_q.positive?
#=> nil
```

PR #611, opened 2022-11-09
Feature/reworks can edit non stock bins

lmatoti:
Reworks
-  For single & bulk bin edit: disallow operations on shipped, tipped or scrapped bins unless user has (new) user_permission: 'can_edit_shipped_bins'
-  For bulk bin edit: search form: added bin_load_id as a search field: So that users can search for bins based on the bin load that they are/were on (bin.bin_load_product_id)


lib/production/repositories/reworks_repo.rb
```diff
@@ -1423,6 +1423,15 @@ def find_bin_ids_by_asset_numbers(bin_asset_numbers)
...
     end
 
+    def find_non_stock_bins(ids)
+      qry = <<~SQL
+        SELECT id
+        FROM rmt_bins
+        WHERE id IN (#{ids.join(',')}) AND (scrapped is true or bin_tipped is true or shipped_asset_number is not null)
+      SQL
+      DB[qry].map(:id)
+    end
```
jcsjcs, line 1433: This coud be simply stated as:
```ruby
ids - DB[:rmt_bins].where(id: ids, shipped_asset_number: nil, bin_tipped: false, scrapped: false).select_map(:id)
```
Otherwise if you're going to use SQL, make use of a positional parameter:
```ruby
      qry = <<~SQL
        SELECT id
        FROM rmt_bins
        WHERE id IN ?
        AND (scrapped is true or bin_tipped is true or shipped_asset_number is not null)
      SQL
      DB[qry, ids].map(:id)
```

jcsjcs (changes requested):

PR #612, opened 2022-11-10
Feature/migration for buildup inspections

WernerVdM97:
Summary of this change:
Added three columns to `pallets` table in preperation for the "buildup inspection" feature.

CHANGELOG entry:

### Added
- `buildup_inspection_required` - bool,
- `built_up_at` - date time, and
- `built_up` - bool to `pallets`.


db/migrations/202211101033_add_buildup_inspection_columns_to_pallets.rb
```diff
@@ -0,0 +1,9 @@
+Sequel.migration do
+  change do
```
jcsjcs, line 2: Implement all migrations with `up` and `down` blocks, not `change` blocks.

jcsjcs (changes requested):

db/migrations/202211101033_add_buildup_inspection_columns_to_pallets.rb
WernerVdM97: Okay, I'll change it accordingly, 

I used `change` because the scaffolding generated it by default.

PR #613, opened 2022-11-10
populate dest country if nil

WernerVdM97:
Summary of this change:
From QA testing we realised that the target country (for a govt sheet where a country isn't specified) was never "auto-filled" on the pallet, contrasting what we previously believed. Therefore, a function has been implemented to specify the first country linked to the chosen eCert agreement as the destination country in the API call to eCert.

CHANGELOG entry:

### Added
- Function `find_first_country_iso_for(agreement_id)` to `ecert_repo`.
### Changed
- If `destination_country` for a `govt_inspection_sheet/-pallet` is nil, then the `TargetCountry` field of the API call to eCert is populated by `find_first_country_iso_for`.


lib/finished_goods/repositories/ecert_repo.rb
```diff
@@ -94,8 +94,14 @@ def find_dest_country(pallet_id)
       DB[qry, pallet_id].get(:iso_country_code)
     end
 
-    def compile_preverify_pallets(pallet_numbers, agreement_code: nil, mixed_tms: false) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
-      return [{}] if pallet_numbers.nil? || (!mixed_tms && agreement_code.nil?)
+    def find_first_country_iso_for(agreement_id)
+      country_ids = get(:ecert_agreements, :country_ids, agreement_id)
+      country_id = country_ids[0]
+      get(:destination_countries, :iso_country_code, country_id)
+    end
```
jcsjcs, line 101: You can do this in one query (and also not crash if `country_ids` is nil).
```ruby
DB[:ecert_agreements]
.left_join(:destination_countries, id: Sequel.lit('ecert_agreements.country_ids[1]'))
.where(Sequel[:ecert_agreements][:id] => agreement_id)
.get(Sequel[:destination_countries][:iso_country_code])
```

jcsjcs (changes requested):

PR #616, opened 2022-11-11
inverse allocation grid: add preview label grid action

miramanz:
Summary of this change:
 add preview label grid action to the inverse allocation grid

### Added
- . preview label grid action to the inverse allocation grid

Production | Runs | List Production Runs | allocate product setups | packing points to product setups | preview label


routes/production/runs.rb
```diff
@@ -676,6 +676,20 @@ class Nspack < Roda
             end
           end
         end
+
+        r.on 'preview_label' do
+          res = interactor.preview_allocation_carton_label_for(id, product_setup_id)
+          if res.success
+            filepath = Tempfile.open([res.instance.fname, '.png'], 'public/tempfiles') do |f|
```
jcsjcs, line 683: Please take the code between `if res.success` and `else` and place it in a new method in `helpers/common_helpers.rb`.
Name the method `show_preview_image` and pass `res` to it. Then replace the code in this action with that call and also check the following routes to see if they can use the same helper:
* `routes/labels/labels.rb`
* `routes/masterfiles/locations.rb`
* `routes/production/runs.rb`

jcsjcs (changes requested):

PR #621, opened 2022-11-17
Feature/buildup inspections

lmatoti:
Inspected Pallet Buildups
- For inspected destination pallet, clone pallet/pallet_sequences/cartons. Then scrap pallet/pallet_sequences/cartons
- Present user with print_pallet_label form


lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -277,6 +277,13 @@ def complete_pallet_buildup(id) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
           end
 
           repo.update(:pallet_buildups, id, updates)
+          new_pallet_id = nil
+          unless pallet_buildup[:auto_create_destination_pallet]
+            res = BuildupInspection.call(dest_pallet_id, @user.user_name)
+            raise Crossbeams::InfoError, res.message unless res.success
+
+            new_pallet_id = res.instance
+          end
```
jcsjcs, line 286: Please implement this more clearly:
```ruby
if buildup_inspection_required?(dest_pallet_id, pallet_buildup[:auto_create_destination_pallet])
  res = PreparePalletForBuildupInspection.call(dest_pallet_id, @user.user_name)
  raise Crossbeams::InfoError, res.message unless res.success

  new_pallet_id = res.instance
end

def buildup_inspection_required?(x,y)
  return false if auto
  return false if not inspected

  true
end
```
So:
* rename the service to `PreparePalletForBuildupInspection`
* and move the inspected/re-inspected check out of the service to a new method.

jcsjcs (changes requested):

lib/production/repositories/reworks_repo.rb
```diff
@@ -277,18 +277,17 @@ def clone_inspected_buildup_pallet_sequences(old_pallet_id, pallet_id, sequence_
...
 
-    def clone_inspected_buildup_cartons(old_sequence_id, new_sequence_id) # rubocop:disable Metrics/AbcSize
-      messcada_repo = MesscadaApp::MesscadaRepo.new
+    def clone_inspected_buildup_cartons(old_sequence_id, new_sequence_id)
       old_carton_label_ids = select_values(:cartons, :carton_label_id, pallet_sequence_id: old_sequence_id)
       old_carton_label_ids.each do |old_carton_label_id|
         carton_label_rejected_fields = %i[id created_at updated_at]
         attrs = find_hash(:carton_labels, old_carton_label_id).reject { |k, _| carton_label_rejected_fields.include?(k) }
-        carton_label_ids = messcada_repo.create_carton_labels(1, attrs)
+        carton_label_ids = MesscadaApp::MesscadaRepo.new.create_carton_labels(1, attrs)
```
jcsjcs, line 285: This can just be:
```ruby
carton_label_id = create(:carton_labels, attrs)
```

jcsjcs (changes requested):

PR #622, opened 2022-11-18
Weighed Fullness- Bin Weighing

lmatoti:
Weighed Fullness- Bin Weighing
1. Negative nett_weight error implemented


lib/messcada/services/bin_weigh.rb
```diff
@@ -124,8 +123,10 @@ def calc_tare_weight
                                             rmt_inner_container_type_id: inputs.rmt_inner_container_type_id)
     end
 
-    def enforce_bin_weight_ranges(bin_fullness, weighed_fullness)
+    def enforce_bin_weight_ranges(bin_fullness, weighed_fullness, max) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
       return ok_response if inputs.presorted_bin || !AppConst::CR_RMT.enforce_bin_weight_ranges?
+      return failed_response("weighed_fullness could not be calculated. nett_weight(#{weights[:nett_weight].to_f}) is out of range, it cannot be smaller than 0.") if !weights[:nett_weight].positive? && !inputs.presorted_bin && AppConst::CR_RMT.enforce_bin_weight_ranges?
+      return failed_response("weighed_fullness could not be calculated. nett_weight(#{weights[:nett_weight].to_f}) is out of range, it cannot be bigger than maximum allowed value of #{max.to_f}.") if weights[:weighed_fullness].nil? && !inputs.presorted_bin && AppConst::CR_RMT.enforce_bin_weight_ranges?
```
jcsjcs, line 129: Please make this easier to read:
```ruby
if AppConst::CR_RMT.enforce_bin_weight_ranges? && !inputs.presorted_bin
  return failed_response("weighed_fullness could not be calculated. nett_weight(#{weights[:nett_weight].to_f}) is out of range, it cannot be smaller than 0.") unless weights[:nett_weight].positive?
  return failed_response # max check
end
```
Also the max check seems to check if the `weighed_fullness` is nil instead of comparing to the `max` value. Is this right?

jcsjcs (changes requested):

PR #623, opened 2022-11-21
Refactor / Orders Target Customers

FritzViljoen:
Summary of this change:

Move target_customer from orders to order_items

CHANGELOG entry:
### Changed
- Move target_customer from orders to order_items

test create order with order items


lib/finished_goods/ui_rules/order_item_rule.rb
```diff
@@ -150,25 +157,7 @@ def make_form_object
...
-                                    inventory_id: nil,
-                                    carton_quantity: nil,
-                                    price_per_carton: nil,
-                                    price_per_kg: nil,
-                                    sell_by_code: nil,
-                                    pallet_format_id: nil,
-                                    pm_mark_id: nil,
-                                    pm_bom_id: nil,
-                                    rmt_class_id: nil)
+      @form_object = OpenStruct.new
```
jcsjcs, line 160: Rather make this:
```ruby
@form_object = new_form_object_from_struct(FinishedGoodsApp::OrderItem)
```
It does not make a material difference to the behaviour, but if the form object needs any default value in future the code will be easier to maintain.

jcsjcs (changes requested):

PR #635, opened 2022-11-29
Pallet Buildups Cloned Destination Pallets

lmatoti:
Pallet Buildups Cloned Destination Pallets
- Log status 'DONATED CARTONS TO {dest pallet}' for src pallets that still have cartons
- Added new column: 'intermediate_pallet'
- Cloned pallet is now stored inside buildups.destination_pallet and the pallet it was cloned from is stored in buildups.intermediate_pallet
- Script to populate buildups.intermediate_pallet and fix buildups.destination_pallet for existing buildups for inspected pallets


scripts/fix_pallet_buildups_intermediate_pallet.rb
```diff
@@ -0,0 +1,69 @@
+# frozen_string_literal: true
```
jcsjcs, line 1: Please re-generate this script from the UI so that it is built from the latest template.

jcsjcs (changes requested):

PR #646, opened 2022-12-02
Feature/central carton verification

WernerVdM97:
Summary of this change:
Added TCP scanners, that are linked to the central MesServer, to the generation of peripherals.xml.
Also added the URL `/messcada/production/central_carton_verification?carton_number=123` to production routes.
This URL acts the same as `/messcada/production/carton_verification`, with the difference that a device parameter is not required.


lib/production/services/build_peripherals_xml.rb
```diff
@@ -48,6 +49,25 @@ def build_xml # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, M
...
+                              NetworkInterface: scanner[:network_interface],
+                              Port: scanner[:port],
+                              TTL: scanner[:ttl],
+                              CycleTime: scanner[:cycle_time],
+                              NoRead: extended_conf['no_read'],
+                              BufferSize: extended_conf['buffer_size'] || 256,
+                              EndOfInput: extended_conf['end_of_input'],
+                              StartOfInput: extended_conf['start_of_input'],
+                              StripEndOfInput: extended_conf['strip_end_of_input'] || 'true',
+                              StripStartOfInput: extended_conf['strip_start_of_input'] || 'true')
```
jcsjcs, line 68: This element will also require `URL: <logic_to_come>` and `Par1: <logic_to_come>`.
The logic is still to be worked out, and there may be more than one `Par` in certain cases.
At this stage it is a good idea to just add placeholders to the XML.

jcsjcs (changes requested):

PR #655, opened 2022-12-08
Feature/depot receipts

WernerVdM97:
[Summary of this change:](https://docs.google.com/document/d/1jO-U0t7zg-4n1FzBjMyaYOY-ux_qRUbDahLHj_ydwQs/edit?usp=sharing)

If the client rule `depot_receipts_require_user_validation` is true, then users are given a new app (based on the existing manual intakes) on which a PO-in file can be manually edited before being accepted and the records generated in the DB.

CHANGELOG entry:

### Added
- For EDI | Receipts | EDI in Txs:
  - New columns, `validated by user` and `validated by user at`
  - New action menu item `manually validate`. Visible if: 
    - `complete` and `newer_edi_received` is false
    - `schema_valid` and `valid` is true
    - user has the `can manually validate depot pallets` permission
 - `Validate depot receipt` page (based on EDI | Actions | manual intake) on which users can delete rows (red), add rows (green) or change rows (orange).
   - Summary page of the changes made, accessible via the `Accept PO as correct` button.
### Fixed
- Improved the handling of manual intakes inline edit grid by adding a hidden `ID` column to each pallet record in the PO-in file.
  -  This allows for the grid to be ordered by pallet number for an improved user experience.
- Validation to prevent a non-number pallet number from being entered.

As Kromco, ensure the user has the required permission.
Find a PO file and receive it. Thereafter, validate it manually via the action menu. The pallet records should not exist.
It is important to test that manual intakes and normal PO-in files (for the client rule set to false) still works as before.


jcsjcs:
> I noticed now that there is no test directory for `lib/edi` nor an `test/routes/edi` directory.
> 
> Are EDI files something we don't usually write tests for?

No. All tests are welcome. The EDIs just don't have any because they didn't fit the usual scaffold shape and the original code was written to a deadline.

WernerVdM97:
> > I noticed now that there is no test directory for `lib/edi` nor an `test/routes/edi` directory.
> > Are EDI files something we don't usually write tests for?
> 
> No. All tests are welcome. The EDIs just don't have any because they didn't fit the usual scaffold shape and the original code was written to a deadline.

Okay, thanks. I'll generate them with the scaffold then. 

Might be worth revisiting these tests in the near future to write some for the rest of the EDI processes.

PR #656, opened 2022-12-08
cartons per pallet masterfile import API

miramanz:
Summary of this change:
added strapper_recipe and securing_sheets as optional values to the cartons per pallet masterfile import API

### Added
- . strapper_recipe and securing_sheets as optional values to the cartons per pallet masterfile import API

*create
curl -v --data '{"masterfile":"cartons_per_pallet","columns":{"pallet_base":"S","pallet_stack_type":"H","basic_pack":"EC250","cartons_per_pallet":123452,"layers_per_pallet":1,"securing_sheets":"1,2","strapper_recipe":"ABC"}}' --header 'Content-Type: application/json' http://localhost:9296/api/masterfiles/create

*update
curl -v --data '{"masterfile":"cartons_per_pallet","keys":{"pallet_base":"S","pallet_stack_type":"H","basic_pack":"EC250","cartons_per_pallet":123452},"columns":{"description":"ABCD descr","strapper_recipe":"ABC"}}' --header 'Content-Type: application/json' http://localhost:9296/api/masterfiles/update

*delete
 curl -v --data '{"masterfile":"cartons_per_pallet","keys":{"pallet_base":"S","pallet_stack_type":"H","basic_pack":"EC250","cartons_per_pallet":123452}}' --header 'Content-Type: application/json' http://localhost:9296/api/masterfiles/delete


lib/masterfiles/repositories/api_repo.rb
```diff
@@ -417,6 +423,14 @@ def update_cartons_per_pallet(keys, attrs) # rubocop:disable Metrics/AbcSize, Me
         attrs.merge!(basic_pack_id: basic_pack_id)
       end
 
+      strapper_recipe = attrs.delete(:strapper_recipe)
```
jcsjcs, line 426: The user could pass in a null strapper recipe to unset it, so if the column is present with a null value, the id should be set to nil.

jcsjcs (changes requested):

lib/masterfiles/repositories/api_repo.rb
```diff
@@ -423,10 +423,10 @@ def update_cartons_per_pallet(keys, attrs) # rubocop:disable Metrics/AbcSize, Me
         attrs.merge!(basic_pack_id: basic_pack_id)
       end
 
-      strapper_recipe = attrs.delete(:strapper_recipe)
-      if strapper_recipe
+      if attrs.include?(:strapper_recipe)
```
jcsjcs, line 426: Please use `key?` instead of `include?` in this case - it reads more obviously.

jcsjcs (changes requested):

PR #660, opened 2022-12-19
Gross Weight Validations

lmatoti:
Gross Weight Validations
Validates that the gross weights provided by users or scales are correct
Affected applications

- palletizing (PDT)
- pallet verification (PDT)
- Repack pallets (PDT)
- /messcada/fg/pallet_weighing
- /messcada/fg/pallet_verification/pallet_weighing/pallet_labeling
- /reworks/single_pallet_edit


lib/production/repositories/reworks_repo.rb
```diff
@@ -248,6 +248,14 @@ def repack_pallet(pallet_id, user_name)
       success_response('ok', new_pallet_id: new_pallet_id)
     end
 
+    def validate_pallet_gross_weight(gross_weight)
+      min = AppConst::CR_PROD.min_pallet_gross_weight
```
jcsjcs, line 252: Return an `ok_response` if the gross weight is nil.

jcsjcs (changes requested):

PR #679, opened 2023-01-18
Hotfix/add packing end date to seasons

WernerVdM97:
Summary of this change:
Changed `CalenderRepo.for_select_seasons` function to filter seasons on new column `packing_end_date`.

CHANGELOG entry:

### Added
- `seasons.packing_end_date` migration, CRUD and grid column.
### Changed
- The function `for_select_seasons` now returns seasons where `packing_end_date` is greater than the current date (instead of previously filtering on `end_date`.

Create/edit any record which has a seasons dropdown, such as production runs, and see if it filter correctly on `packing_end_date`
`packing_end_date` should default to `end_date` if left blank in seasons masterfile.


db/migrations/202301181406_add_packing_end_date_to_seasons.rb
```diff
@@ -0,0 +1,16 @@
+Sequel.migration do
+  up do
+    alter_table(:seasons) { add_column :packing_end_date, Date }
```
jcsjcs, line 3: Change this to use do/end blocks, please - keep consistent in migrations.

db/migrations/202301181406_add_packing_end_date_to_seasons.rb
```diff
@@ -0,0 +1,16 @@
...
+
+    run 'UPDATE
+          seasons
+        SET
+          packing_end_date=end_date
+        WHERE
+          packing_end_date IS NULL'
+  end
```
jcsjcs, line 11: Make the column NOT NULL after the update.

jcsjcs (changes requested):

db/migrations/202301181406_add_packing_end_date_to_seasons.rb
```diff
@@ -0,0 +1,22 @@
...
+    end
+
+    run 'UPDATE
+          seasons
+        SET
+          packing_end_date=end_date
+        WHERE
+          packing_end_date IS NULL'
+
+    run 'ALTER TABLE seasons ALTER COLUMN packing_end_date SET NOT NULL;'
```
jcsjcs, line 14: Rather use the Sequel DSL for this part - `alter_table`, `set_column_not_null`.

jcsjcs (changes requested):

PR #691, opened 2023-02-01
correct default datatypes

PatieC:
Summary of this change:

Change column data types

CHANGELOG entry:

 .
### Changed
Change column data types

Dataminer | Reports | List  Reports


reports/incentive.yml
```diff
@@ -278,7 +278,7 @@
     :sequence_no: 9
     :caption: Total fruit
     :namespaced_name: 
-    :data_type: :string
+    :data_type: :numeric
```
jcsjcs, line 281: Did you use the Datamine | Manage | Report Admin menu to do this?

- `numeric` should be `number`.
- Also the format should be delimited.

reports/incentive_report_per_comm_and_pack.yml
```diff
@@ -12,10 +12,18 @@
           basic_pack_codes.basic_pack_code as pack_code,
           day_night_or_custom as shift_type_code,
           count(cartons.id) AS total_cartons,
+   wage_levels.wage_level ,
```
jcsjcs, line 15: Indentation is wrong.

jcsjcs (changes requested):

reports/incentive.yml
PatieC: I will correct, I was doing it manually in the file.

PatieC: Does this apply to all the fields  I have changed to numeric

jcsjcs: Yes. Use the Dataminer function.

PR #692, opened 2023-02-01
Hotfix/add default po in pallet sequence mass

WernerVdM97:
**Summary of this change:**
Added a client rule to specify the default value in PO-in files for pallet sequences with a zero weight.

CHANGELOG entry:

### Added
- Client rule `default_depot_pallet_sequence_nett_weight` with value `nil` for all clients, except for `um` which is `111`.

Receive a PO-in file with `mass = 0` for a given sequence and see if the `pallet_sequence.nett_weight` becomes equal to `111` when accepted.
For any other client, the PO-in should still fail and specifiy that the mass cannot be zero.


lib/client_rules/client_edi_rules.rb
```diff
@@ -267,6 +277,12 @@ def exclude_partial_instock_pallets_from_ps?(explain: false)
       setting(:exclude_partial_instock_pallets_from_ps)
     end
 
+    def default_depot_pallet_sequence_nett_weight(explain: false)
+      return 'The default value to replace a null nett weight for a pallet sequence in incoming PO files.' if explain
```
jcsjcs, line 281: Null or zero.

jcsjcs (changes requested):

PR #694, opened 2023-02-02
Feature/carton equals pallet

miramanz:
Summary of this change:
add dialogue to set and/or unset carton equals pallet for lines and their child robots.

### Added
- . dialogue to set and/or unset carton equals pallet for lines and their child robots.

Production | Resources | Plant Resources | set_carton_equals_pallet
Production | Resources | Plant Resources | unset_carton_equals_pallet


lib/production/repositories/resource_repo.rb
```diff
@@ -336,6 +336,12 @@ def enable_disable_plant_resource(id, enable: true)
       DB[:system_resources].where(id: DB[:plant_resources].where(id: id).get(:system_resource_id)).update(active: enable)
     end
 
+    def set_unset_carton_equals_pallet(plant_resource_id, enable: true)
+      attrs = { resource_properties: { carton_equals_pallet: enable ? 't' : 'f' } }
```
jcsjcs, line 340: The update needs to be respectful of any other properties in `resource_properties` (should there be any in future), so you need to do something like this:
```ruby
bool = enable ? 't' : 'f'
query = %(UPDATE plant_resources SET resource_properties = CASE WHEN resource_properties IS NULL THEN '{"carton_equals_pallet": "#{bool}"}' ELSE resource_properties || '{"carton_equals_pallet": "#{bool}"}' END WHERE id IN ?)
DB[query, ids].update
```

jcsjcs (changes requested):

PR #708, opened 2023-02-23
Prepend bin asset_numbers

miramanz:
Summary of this change:
Prepend bin asset_numbers with a fixed string, from client rule- if set, wherever bins and rebins are created
* This affects all places where rebins are created: PDT, Production_runs, rebin_verification
* Also affected is: preprint bin labels and new delivery{auto_allocate_bin_numbers}
* Please implement rebin first, then create 2nd feature for orchard run (rmt) bins

### Added
- . bin_asset_number_prefix production run client rule (Only relevant to MC (Mouton); value for them will be 'B')
### Changed
- . the get_available_bin_asset_numbers method to prepend bin asset_numbers with client rule value
-  a number of places to accept bin asset number as string not number 

*create rebin
RMD | Raw Materials | Create Rebin
Production | Runs | rebins | create rebin

*rebin verification
RMD | Raw Materials | Rebin Verification
robot mock - /messcada/production/rebin_verification

*preprint labels
Production | InProgress | Active Setups | Print Carton Label

* auto allocate bin numbers
RMD | Deliveries | edit | auto allocate bin numbers

*edit bin
RMD | Raw Materials | Edit Bin

Changed to accept bin asset number as string not number 
*bin enquiry
RMD | Raw Materials | Bin Enquiry

RMD | Raw Materials | Store orchard bins
RMD | Raw Materials | Store presorted bins


lib/raw_materials/repositories/rmt_delivery_repo.rb
```diff
@@ -328,18 +328,20 @@ def rebin_details(id)
...
           seq = DB["SELECT nextval('bin_asset_seq')"].get
-          id = DB[:bin_asset_numbers].insert(bin_asset_number: seq.to_s)
-          ar << [id, seq.to_s]
+          bin_asset_number = AppConst::CR_PROD.bin_asset_number_prefix.empty? ? seq.to_s : "#{AppConst::CR_PROD.bin_asset_number_prefix}#{seq}"
+          id = DB[:bin_asset_numbers].insert(bin_asset_number: bin_asset_number)
+          ar << [id, bin_asset_number]
         end
         ar
       else
+        bin_asset_number = AppConst::CR_PROD.bin_asset_number_prefix.empty? ? 'a.bin_asset_number' : "'#{AppConst::CR_PROD.bin_asset_number_prefix}' || a.bin_asset_number"
```
jcsjcs, line 342: Surely the bin assets table will contain already-prefixed numbers when appropriate?
In which case this query does not need to change.

lib/raw_materials/interactors/rmt_bin_interactor.rb
```diff
@@ -703,8 +703,9 @@ def delete_rmt_bin(id)
       failed_response(e.message)
     end
 
-    def validate_is_rebin(id, bin_number)
+    def validate_is_rebin_or_presorted(id, bin_number)
       return failed_response("Rebin Scanned:#{bin_number}. Please scan a bin instead.") unless repo.exists?(:rmt_bins, id: id, production_run_rebin_id: nil)
+      return failed_response("Presort Bin Scanned:#{bin_number}. Please scan a bin instead.") if repo.get(:rmt_bins, :presorted, id)
```
jcsjcs, line 708: These rules should be handled by a single call to the RMT bin's `TaskPermissionCheck` with a new check named `is_orchard_bin`.

lib/raw_materials/repositories/rmt_delivery_repo.rb
jcsjcs: This might be handled by adding an extra condition to the where clause (if there is a prefix rule):
```SQL
AND a.bin_asset_number LIKE '#{prefix}%'
```
This will ensure that old non-prefixed assets are never re-used.

jcsjcs (changes requested):

PR #711, opened 2023-02-28
Create Rebin

lmatoti:
Create Rebin
- Show validation errors for rebin fields that are not part of the form.


lib/raw_materials/interactors/rmt_bin_interactor.rb
```diff
@@ -412,7 +415,7 @@ def create_rebin(params) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCo
       instance = rmt_bin(id)
       success_response('Created RMT rebin', instance)
     rescue Sequel::UniqueConstraintViolation
-      validation_failed_response(OpenStruct.new(messages: { status: ['This RMT bin already exists'] }))
+      failed_response('This RMT bin already exists')
```
jcsjcs, line 418: Why not use validation_failed_message_response('This RMT bin already exists')?

jcsjcs (changes requested):

PR #713, opened 2023-03-02
Pallet Buildup

lmatoti:
Pallet Buildup
- Do not perform buildup inpsection if any source pallet has not been inspected


lib/finished_goods/interactors/buildups_interactor.rb
```diff
@@ -313,10 +313,16 @@ def complete_pallet_buildup(id) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
       failed_response(e.message)
     end
 
-    def buildup_inspection_required?(pallet_id, auto_create_destination_pallet)
+    def get_unispected_pallets(id)
```
jcsjcs, line 316: Make the intention more obvious:
`def uninspected_source_pallets_on_buildup(pallet_buildup_id)`

jcsjcs (changes requested):

PR #715, opened 2023-03-06
Bins/Pallets Grids Where Collumn

lmatoti:
Bins/Pallets Grids Where Collumn
- Added a calculated 'where' column to all bins/pallets grids


db/ddl/functions/fn_location_where/202303031045_fn_location_where.sql
```diff
@@ -0,0 +1,20 @@
+CREATE OR REPLACE FUNCTION public.fn_location_where(location_id integer, tripsheet_destination_location_id integer)
```
jcsjcs, line 1: Pass in the text of each location, not the ids. It is very expensive to run two queries for every row returned.

db/migrations/202303061118_add_location_where_to_vw_pallet_sequence_flat.rb
```diff
@@ -0,0 +1,17 @@
+Sequel.migration do
```
jcsjcs, line 1: Put all three migrations in one file.

grid_definitions/dataminer_queries/all_pallets.yml
```diff
@@ -196,6 +196,7 @@
       pallets.nett_weight_externally_calculated,
       pallets.flag,
       pallets.buildup_inspection_required,
+      fn_location_where(pallets.location_id, pallets.tripsheet_destination_location_id) as where,
```
jcsjcs, line 199: `where` is a reserved word.
Change all these queries to return the column as `where_located` and display the column header as `Where`.

jcsjcs (changes requested):

grid_definitions/dataminer_queries/all_pallets.yml
```diff
@@ -196,7 +196,7 @@
       pallets.nett_weight_externally_calculated,
       pallets.flag,
       pallets.buildup_inspection_required,
-      fn_location_where(pallets.location_id, pallets.tripsheet_destination_location_id) as where,
+      fn_location_where(MAX(locations.location_long_code), MAX(tripsheet_destination_location.location_long_code)) as where_located,
```
jcsjcs, line 199: Why do you have MAX for some of these queries? Surely there should only be one location per record?

jcsjcs (changes requested):

PR #722, opened 2023-03-14
Feature/titan api schema

WernerVdM97:
Summary of this change:
Added schema validation to the outgoing inspection request of the TITAN API.
As well as fixed a bug where the custom country name, instead of the ISO code, was sent through to TITAN.

Run TITAN mock API and request inspection for any given consignment that already had a previous request sent.
Compare the JSON requests before and after this update to ensure they are still the same.


lib/finished_goods/repositories/titan_repo.rb
```diff
@@ -212,7 +212,7 @@ def titan_pallet_header(pallet_id, pallet, govt_inspection_sheet, destination_re
         targetCountry: destination_country || govt_inspection_sheet.iso_country_code || EcertRepo.new.find_first_country_iso_for(ecert_agreement_id),
         protocolExceptionIndicator: titan_protocol_exception || govt_inspection_sheet.titan_protocol_exception || 'NA',
         agreementCode: ecert_agreement_code,
-        upn: pallet.upn }
+        upn: pallet.upn || '' }
```
jcsjcs, line 215: Is this necessary? You seem to do this again in the validation method.

jcsjcs (changes requested):

lib/finished_goods/repositories/titan_repo.rb
WernerVdM97: In the case of `validate_titan_inspection_api_payload` no it is not necessary, however, I added that to have this function match the convention of the function `titan_pallet_sequence:232` where `phytoData` is made an empty string if nil.

jcsjcs: OK. But aren't these methods only used to build up a hash for input to the dry validation?

In which case whether you pass in `''` or `nil` as a value, the output from the validation will always be `nil`.
And in that case, isn't `pre-validation-value = value || ''` just noise?

jcsjcs (changes requested):

lib/finished_goods/repositories/titan_repo.rb
WernerVdM97: Yes, it is just noise. I'll remove both cases then.

PR #750, opened 2023-04-13
Disallow Storage Of Unweighed Rebins

lmatoti:
Disallow Storage Of Unweighed Rebins
- This validation is done at the begining of the process for all putaway apps


lib/raw_materials/repositories/rmt_delivery_repo.rb
```diff
@@ -980,5 +980,14 @@ def calc_converted_bin_gross_weight_from_pallet(pallet_id) # rubocop:disable Met
 
       plt_nett_weight + material_mass + plt_std_pack_material_mass
     end
+
+    def validate_move_unweighed_rebins(bin_numbers)
```
jcsjcs, line 984: This would be **much** more efficient if you wrote it as:
```ruby
def validate_move_unweighed_rebins(bin_numbers)
  return ok_response unless AppConst::CR_RMT.block_moves_of_unweighed_rebins?

  DB[:rmt_bins]
    .where(bin_asset_number: bin_numbers)
    .exclude(production_run_rebin_id: nil)
    .select(:gross_weight, :bin_asset_number).each do |bin|
    return failed_response("Rebin (#{bin[:bin_asset_number]}) must be weighed before it can be stored in a location") if bin[:gross_weight].to_f.zero?
  end

  ok_response
end
```

jcsjcs (changes requested):

lib/finished_goods/services/move_stock.rb
```diff
@@ -65,6 +65,11 @@ def call # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metric
       success_response("#{stock_type}: #{stock_item_number} moved successfully", upd)
     end
 
+    def self.validate_block_move_of_unweighed_rebin(stock_item_id)
```
jcsjcs, line 68: This is not really part of the MoveStock service.
Implement it as an RMT bin task permission check: `can_bin_be_moved`.

jcsjcs (changes requested):

PR #762, opened 2023-04-27
Hotfix/batch preprint rebins fix

WernerVdM97:
Summary of this change:
Bugfix to accommodate clients who batch preprint rebin labels and have the client rule `use_db_sequence_for_bin_asset_numbers` set to true.

CHANGELOG entry:

### Fixed
- Batch pre-printing a rebin label no longer causes `adding a duplicate` validation error.

As a client with the client rule set to true go to Production | In Progress | active setups and try to print rebin labels. Be sure to try both singular as well as multiple and confirm that it still works as intended for clients who have the client rule set to false.


lib/messcada/services/batch_print_carton_labels.rb
```diff
@@ -33,20 +33,25 @@ def call # rubocop:disable Metrics/AbcSize
...
         raise Crossbeams::InfoError, res.message unless res.success
-
-        attrs[:rebin_asset_number] = res.instance[:bin_asset_number]
       end
       repo.transaction do
-        ids = repo.create_carton_labels(no_of_prints, attrs)
+        ids = nil
+        if allocate_rebin_asset_numbers
+          res.instance[:bin_asset_numbers].each do |bin_asset_number|
+            ids = repo.create_carton_labels(1, attrs.merge(rebin_asset_number: bin_asset_number))
```
jcsjcs, line 43: This will continually rewrite the `ids` var with the last label created, so it will only print one label.

jcsjcs (changes requested):

PR #787, opened 2023-06-05
Feature/max weight constraint bin tipping

lmatoti:
Production Bin Tipping
- Implemented maximum weight constraint on gross_weight field


jcsjcs:
This PR includes changes from another PR. The single change to `lib/production/ui_rules/tip_bin_rule.rb` has been made in a commit on develop instead.

PR #809, opened 2023-06-26
Orders With Missing Prices

lmatoti:
Orders With Missing Prices
- On Manage Order : Notify the user if the order has order_items with missing prices.
- On Close Order : Notify the user if the order has order_items with missing prices.
- Added calculated field 'Missing Prices' on orders grid, colour the row orange if missing prices


lib/finished_goods/repositories/order_repo.rb
```diff
@@ -308,5 +308,11 @@ def apply_historical_order_prices(order_id)
         update(:order_items, order_item_id, params)
       end
     end
+
+    def missing_prices_line_items(order_id)
+      DB[:order_items]
+        .where(Sequel.lit("order_id=#{order_id} and (price_per_carton IS NULL OR price_per_kg IS NULL)"))
```
jcsjcs, line 314: Shouldn't this be an `AND`, not an `OR`:
```ruby
DB[:order_items]
.where(order_id: order_id, price_per_carton: nil, price_per_kg: nil)
.count
.positive?
```

jcsjcs (changes requested):

lib/finished_goods/repositories/order_repo.rb
lmatoti: This is what Hans specified:

1) On pressing 'close' button on an order: If any order_item has a null or zero value for 'price_per_carton' or 'price_per_kg': Show a message (notice presumably) 'Order has missing prices for <n> line item(s)'

jcsjcs: OK, thanks. I understood incorrectly that only one or the other must exist.

